PR #2, opened 2023-09-19
cmake: Add support to pass SANITIZE argument

harishdm:
cmake builds now support passing -DSANITIZE argument to enable various sanitizers.

Also, add README.md with instructions on how to build fuzzers


harishdm:
@DichenZhang1 @ram-mohan can you please review? Thanks.

ram-mohan (approved):

PR #8, opened 2023-10-02
cmake.yml: Test cmake builds using gcc

harishdm:
cmake configure step was wrongly hard-coded to clang which prevented testing the gcc based builds. 
It is now updated to pick appropriate compilers based on the build type.


harishdm:
@DichenZhang1 I have update cmake.yml to ensure that gcc builds are validated correctly.
Can you please review?
Thanks.

PR #10, opened 2023-10-02
cmake: Pass -DENABLE_FUZZERS=ON to enable building fuzzers

harishdm:
Fuzzer build is now based on an argument passed to cmake. 
With this, macos builds can now be verified for each pull request.


harishdm:
@DichenZhang1 with this cmake builds on macos will also be verified.
Can you please review?
Thanks.

Issue #7, opened 2023-09-29
Error when building


jeremysong1106:
Hi, 
I tried to build the code on my Ubuntu vm, and it said that the third party library doesn't exist, and I checked the folder again, the googletest and libjpeg-turbo is not pull from git.
I tried to manually pull the third party library and build it , then the error shows up :

undefined reference to `pthread_setspecific'
collect2: error: ld returned 1 exit status

am I building the code correctly? I'm looking forward to you helps,

Bests

ram-mohan:
target_link_libraries(ultrahdr_unit_test
    ${SRC_DIR}/third_party/build/googletest/src/googletest-build/lib/libgmock.a
    ${SRC_DIR}/third_party/build/googletest/src/googletest-build/lib/libgtest.a
    ${SRC_DIR}/third_party/build/googletest/src/googletest-build/lib/libgtest_main.a
    pthread
)

can you add pthread to the list of libraries and see if it resolves the issue

harishdm:
Hi @jeremysong1106, 
Thank you for reporting this issue. 
pthread dependency has been added in the latest commit. Can you please try with the latest commit? 

Also, for the issue with googletest and libjpeg-turbo not getting cloned in your VM setup, can you please confirm if your VM has git (version 1.6.5 or later) installed on it?

PR #11, opened 2023-10-03
cmake: Add pthread dependendency

harishdm:
When using some older compilers, the missing pthread dependency results in build errors.


harishdm:
@DichenZhang1 this fixes the pthread related failure in #7 seen in some older versions of the compilers.

PR #16, opened 2023-10-11
Fix compilation issue while building with logs enabled

ram-mohan:


include/ultrahdr/ultrahdrcommon.h
```diff
@@ -25,11 +25,31 @@
...
-        #define ALOGD(...) fprintf(stderr, __VA_ARGS__)
-        #define ALOGE(...) fprintf(stderr, __VA_ARGS__)
-        #define ALOGI(...) fprintf(stdout, __VA_ARGS__)
-        #define ALOGV(...) fprintf(stdout, __VA_ARGS__)
-        #define ALOGW(...) fprintf(stderr, __VA_ARGS__)
+        #define ALOGD(...)                \
+          do {                            \
+            fprintf(stderr, __VA_ARGS__); \
+            fprintf(stderr, "\n");        \
+          } while (0);
```
DichenZhang1, line 32: I don't think the semicolon is necessary here.

harishdm: Thanks for flagging this. Removed in the next commit in this pull request.

DichenZhang1 (approved):

PR #19, opened 2023-10-12
add profiling code for encode and decode api calls

ram-mohan:


tests/ultrahdr_app.cpp
```diff
@@ -62,6 +63,27 @@ const float BT2020RGBtoYUVMatrix[9] = {0.2627,
                                        (-0.6780 / 1.4746),
                                        (-0.0593 / 1.4746)};
 
+const int kProfileCount = 10;
```
DichenZhang1, line 66: this line can be protected by "#ifdef PROFILE_ENABLE"

DichenZhang1 (approved):

PR #22, opened 2023-10-18
Fuzz

ram-mohan:


harishdm (approved):

include/ultrahdr/gainmapmath.h
```diff
@@ -136,8 +136,8 @@ inline uint16_t floatToHalf(float f) {
   // round-to-nearest-even: add last bit after truncated mantissa
   const uint32_t b = *((uint32_t*)&f) + 0x00001000;
 
-  const uint32_t e = (b & 0x7F800000) >> 23; // exponent
-  const uint32_t m = b & 0x007FFFFF; // mantissa
+  const int32_t e = (b & 0x7F800000) >> 23; // exponent
```
DichenZhang1, line 139: Why this change is needed? Per the calculation here both `e` and `m` should always be positive.

DichenZhang1 (approved):

DichenZhang1 (approved):

DichenZhang1 (approved):

include/ultrahdr/gainmapmath.h
harishdm: I have moved this to a different branch to unblock the first three CLs. Will follow up this in a separate pull request.

ram-mohan: operations ((e - 112) and (125 - e) in the return statement below is causing ubsan issue.

PR #24, opened 2023-10-19
Allocate width aligned to MCU size for boundary rows

ram-mohan:
jpeg_write_raw_data() expects input size to be aligned to MCU. If the image dimensions are not aligned then padding is done for boundary rows. The padded rows allocation is not proper and this is corrected.

Bug: oss-fuzz-63314
Test: ultrahdr_enc_fuzzer

### DIFF
--- a/jpegencoderhelper.cpp
+++ b/jpegencoderhelper.cpp
@@ -147,15 +147,15 @@ bool JpegEncoderHelper::compressYuv(jpeg_compress_struct* cinfo, const uint8_t* 
     JSAMPROW cr[kCompressBatchSize / 2];
     JSAMPARRAY planes[3]{y, cb, cr};
 
+    const int aligned_width = ALIGNM(cinfo->image_width, kCompressBatchSize);
+    const bool need_padding = (lumaStride < aligned_width);
     size_t u_plane_size = chromaStride * cinfo->image_height / 2;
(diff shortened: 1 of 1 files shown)


DichenZhang1 (approved):

jpegencoderhelper.cpp
```diff
@@ -142,46 +142,58 @@ void JpegEncoderHelper::setJpegCompressStruct(int width, int height, int quality
 
 bool JpegEncoderHelper::compressYuv(jpeg_compress_struct* cinfo, const uint8_t* yBuffer,
                                     const uint8_t* uvBuffer, int lumaStride, int chromaStride) {
+    size_t u_plane_size = chromaStride * cinfo->image_height / 2;
+    uint8_t* y_plane = const_cast<uint8_t*>(yBuffer);
+    uint8_t* u_plane = const_cast<uint8_t*>(uvBuffer);
+    uint8_t* v_plane = const_cast<uint8_t*>(u_plane + u_plane_size);
+
+    const int aligned_width = ALIGNM(cinfo->image_width, kCompressBatchSize);
```
DichenZhang1, line 150: Nit: maybe renaming it to `aligned_luma_width` to be more clear.

ram-mohan: Updated patchset to resolve this.

harishdm (approved):

PR #26, opened 2023-10-22
Cleanup CMakeLists.txt and Update README.md

ram-mohan:


CMakeLists.txt
```diff
@@ -14,135 +14,184 @@
...
+      ${SRC_DIR}/third_party/googletest/googletest/include
+      ${SRC_DIR}/third_party/googletest/googlemock/include PARENT_SCOPE)
+  set(GTEST_BOTH_LIBRARIES
+      ${SRC_DIR}/third_party/build/googletest/src/googletest-build/lib/libgtest.a
+      ${SRC_DIR}/third_party/build/googletest/src/googletest-build/lib/libgtest_main.a PARENT_SCOPE)
+endfunction()
+
+fetch_libjpegturbo()
+
+if(${ENABLE_FUZZERS})
```
harishdm, line 111: GTest is independent of Fuzzers.
Can we just use the else part?

harishdm (approved):

README.md
```diff
@@ -1,30 +1,71 @@
...
-```
+Background
+==========
+
+libultrahdr is an image compression library that uses gain map technology
+to store and distribute HDR images. Conceptually on the encoding side, the
+library accepts SDR and HDR rendition of an image and from these a Gain Map
+(quotient between the two renditions) is computed. The library then uses
+backward compatible means to store the base image (SDR), gain map image and
+some associated metadata. Legacy readers that dont support parsing the
```
DichenZhang1, line 9: don't

DichenZhang1 (approved):

PR #45, opened 2023-11-20
Add benchmark tests for encoder apis

aayushsoni111:
Change-Id: I972856d603782d3167f7d002896ce87161292d96

### DIFF
--- a/benchmark/benchmark_test.cpp
+++ b/benchmark/benchmark_test.cpp
@@ -23,6 +23,8 @@
 
 using namespace ultrahdr;
 
+std::string kTestImagesPath = "./data/UltrahdrBenchmarkTestRes-1.0/";
+
 std::vector<std::string> kDecodeAPITestImages{
     // 12mp test vectors
     "mountains.jpg",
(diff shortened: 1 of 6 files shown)


harishdm (approved):

benchmark/benchmark_test.cpp
```diff
@@ -23,13 +23,64 @@
...
 
-const std::string kDecTestVectors[] = {
-    "./data/city_night.jpg",    "./data/desert_sunset.jpg", "./data/lamps.jpg",
-    "./data/mountains.jpg",     "./data/desert_wanda.jpg",  "./data/grand_canyon.jpg",
-    "./data/mountain_lake.jpg",
+std::vector<std::string> kDecodeAPITestImages{
+    // 12mp test vectors
+    "mountains.jpg",
+    "mountain_lake.jpg",
+    "desert_wanda.jpg",
```
DichenZhang1, line 30: Are these images checked in?

benchmark/benchmark_test.cpp
```diff
@@ -46,6 +97,34 @@ std::string ofToString(const ultrahdr_output_format of) {
   }
 }
 
+std::string cgToString(const ultrahdr_color_gamut cg) {
```
DichenZhang1, line 100: I'd prefer using `colorGamut` because `cg` is too ambiguous.

benchmark/benchmark_test.cpp
harishdm: these aren't checked in yet. Currently these will need to be manually placed. Once we get a confirmation, we will upload a zip that will contain these files.

DichenZhang1 (approved):

benchmark/benchmark_test.cpp
DichenZhang1: These images are shareable.

harishdm (approved):

DichenZhang1 (approved):

benchmark/benchmark_test.cpp
DichenZhang1: Hi Harish, could you confirm if the images are going to be checked in together with this CL?

harishdm: @DichenZhang1 Images are uploaded to google cloud and the third commit in this pull request downloads the zip and uses those files.

PR #56, opened 2023-12-31
handle xmp packet Header and Trailer during parsing

ram-mohan:
libimage_io fails to handle xmp packet header and trailer. Before handing the data to lib strip these fields

fixes google#54


lib/src/jpegrutils.cpp
```diff
@@ -451,7 +451,35 @@ bool getMetadataFromXMP(uint8_t* xmp_data, size_t xmp_size, ultrahdr_metadata_st
   xmp_size -= nameSpace.size() + 1;
   XMPXmlHandler handler;
 
-  // We need to remove tail data until the closing tag. Otherwise parser will throw an error.
+  // xml parser fails to prase packet header, wrapper. remove them before handing the data to
```
DichenZhang1, line 454: parse

DichenZhang1 (approved):

PR #68, opened 2024-03-04
Replace divides with multiplies in get*Pixel functions

cmacdonald-arm:
Replaces the divides in the aforementioned functions with reciprocal multiplies. This change requires modification of the test functions as well, otherwise the results end up off by one.

This improves performance across the board with the improvement most notable on small in-order Arm cores. This benefit should be amplified by subsequent Neon optimizations.


jwright-arm:
@DichenZhang1 this change would unlock quite a bit of Neon/SIMD performance.

DichenZhang1:
Sure, thank you for your optimization! @jwright-arm

PR #72, opened 2024-03-07
Fixes

ram-mohan:


lib/src/jpegr.cpp
```diff
@@ -665,64 +667,64 @@ status_t JpegR::decodeJPEGR(jr_compressed_ptr jpegr_image_ptr, jr_uncompressed_p
...
-    }
     if (exif->length < jpeg_dec_obj_yuv420.getEXIFSize()) {
       return ERROR_JPEGR_BUFFER_TOO_SMALL;
     }
     memcpy(exif->data, jpeg_dec_obj_yuv420.getEXIFPtr(), jpeg_dec_obj_yuv420.getEXIFSize());
     exif->length = jpeg_dec_obj_yuv420.getEXIFSize();
   }
 
+  JpegDecoderHelper jpeg_dec_obj_gm;
+  jpegr_uncompressed_struct gainmap_image;
```
DichenZhang1, line 678: Reason to move this before `if (output_format == ULTRAHDR_OUTPUT_SDR)`?

if output format is SDR, we don't need to decode the gain map. So I think it's better to run that first, and early terminate.

harishdm: Dichen,
Checks at line 679 and 701 will ensure that these blocks don't execute when not really needed.

DichenZhang1 (approved):

PR #97, opened 2024-04-04
Let library be built as both static and shared

kmilos:
You may define `BUILD_SHARED_LIBS` to FALSE if you really want the STATIC to be the default.


kmilos:
Sorry, not interested in signing a CLA for a trivial change like this, please feel free to make it yourselves accordingly.

ram-mohan:
This is not the correct place for static or shared choice. Because core target is not the final uhdr library. Preferred location is add_library(${UHDR_TARGET_NAME} SHARED). We will enable both static and shared library support.

kmilos:
Ok, thanks. Closing then.

PR #110, opened 2024-04-16
Add support for rgb formats during encode

ram-mohan:
Fixed issues in color space conversion modules of
sample app. Also rename variables in sample app
accurately

Test: ultrahdr_app -m 0 -p cosmat_1920x1080_p010.yuv \
  -w 1920 -h 1080 -q 97 -a 0

Change-Id: I86eeef02f5bc0e6b8c28de0da46b0f6d7eec6358

### DIFF
--- a/lib/include/ultrahdr/gainmapmath.h
+++ b/lib/include/ultrahdr/gainmapmath.h
@@ -19,7 +19,10 @@
 
 #include <array>
 #include <cmath>
+#include <cstring>
 
+#include "ultrahdr_api.h"
+#include "ultrahdr/ultrahdrcommon.h"
 #include "ultrahdr/ultrahdr.h"
 #include "ultrahdr/jpegr.h"
 
@@ -516,6 +519,11 @@ uint32_t colorToRgba1010102(Color e_gamma);
  * Alpha always set to 1.0.
  */
 uint64_t colorToRgbaF16(Color e_gamma);
+
+/*
+ * Helper for preparing encoder raw inputs for encoding
+ */
+std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_t* src);
 
 }  // namespace ultrahdr
 
--- a/lib/src/editorhelper.cpp
+++ b/lib/src/editorhelper.cpp
@@ -331,4 +331,5 @@ std::unique_ptr<uhdr_raw_image_ext_t> apply_resize(ultrahdr::uhdr_resize_effect_
   }
   return std::move(dst);
 }
+
 }  // namespace ultrahdr
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -758,4 +758,183 @@ uint64_t colorToRgbaF16(Color e_gamma) {
          (((uint64_t)floatToHalf(e_gamma.b)) << 32) | (((uint64_t)floatToHalf(1.0f)) << 48);
 }
 
+std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_t* src) {
+  std::unique_ptr<uhdr_raw_image_ext_t> dst = nullptr;
+  Color (*rgbToyuv)(Color) = nullptr;
+
+  if (src->fmt == UHDR_IMG_FMT_32bppRGBA1010102 || src->fmt == UHDR_IMG_FMT_32bppRGBA8888) {
+    if (src->cg == UHDR_CG_BT_709) {
+      rgbToyuv = srgbRgbToYuv;
+    } else if (src->cg == UHDR_CG_BT_2100) {
+      rgbToyuv = bt2100RgbToYuv;
+    } else if (src->cg == UHDR_CG_DISPLAY_P3) {
+      rgbToyuv = p3RgbToYuv;
+    } else {
+      return dst;
+    }
+  }
+
+  if (src->fmt == UHDR_IMG_FMT_32bppRGBA1010102) {
+    dst = std::make_unique<uhdr_raw_image_ext_t>(UHDR_IMG_FMT_24bppYCbCrP010, src->cg, src->ct,
+                                                 UHDR_CR_LIMITED_RANGE, src->w, src->h, 64);
+
+    uint32_t* rgbData = static_cast<uint32_t*>(src->planes[UHDR_PLANE_PACKED]);
+    unsigned int srcStride = src->stride[UHDR_PLANE_PACKED];
+
+    uint16_t* yData = static_cast<uint16_t*>(dst->planes[UHDR_PLANE_Y]);
+    uint16_t* uData = static_cast<uint16_t*>(dst->planes[UHDR_PLANE_UV]);
+    uint16_t* vData = uData + 1;
+
+    for (size_t i = 0; i < dst->h; i += 2) {
+      for (size_t j = 0; j < dst->w; j += 2) {
+        Color pixel[4];
+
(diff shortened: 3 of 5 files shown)


lib/src/editorhelper.cpp
```diff
@@ -234,4 +235,255 @@ std::unique_ptr<uhdr_raw_image_ext_t> apply_resize(uhdr_raw_image_t* src, int ds
...
+        y2 = CLIP3(y2, 64.0f, 940.0f);
+        yData[dst->stride[UHDR_PLANE_Y] * (i + 1) + j] = uint16_t(y2) << 6;
+
+        y3 = (y3 * 876.0f) + 64.0f + 0.5f;
+        y3 = CLIP3(y3, 64.0f, 940.0f);
+        yData[dst->stride[UHDR_PLANE_Y] * (i + 1) + j + 1] = uint16_t(y3) << 6;
+
+        u0 = (u0 + u1 + u2 + u3) / 4;
+        v0 = (v0 + v1 + v2 + v3) / 4;
+
```
DichenZhang1, line 339: Better to add a note "this is narrow-range"

lib/src/editorhelper.cpp
```diff
@@ -234,4 +235,255 @@ std::unique_ptr<uhdr_raw_image_ext_t> apply_resize(uhdr_raw_image_t* src, int ds
...
+        v0 = v0 * 255.0f + 0.5 + 128.0f;
+
+        u0 = CLIP3(u0, 0.0f, 255.0f);
+        v0 = CLIP3(v0, 0.0f, 255.0f);
+
+        uData[dst->stride[UHDR_PLANE_U] * (i / 2) + (j / 2)] = uint8_t(u0);
+        vData[dst->stride[UHDR_PLANE_V] * (i / 2) + (j / 2)] = uint8_t(v0);
+      }
+    }
+  } else if (src->fmt == UHDR_IMG_FMT_12bppYCbCr420) {
```
DichenZhang1, line 437: Question why do we need that?

Ideally I'm thinking this method is converting RGB input to YUV output, and if user calls this method with YUV input, we don't need to do anything but tell the user this is wrong call. With this code I think it will potentially grow into a bigger TODO like YUV 444 transfer to YUV 420, and we don't have support for that currently. So I would suggest remove this and the following if clause, just keep RGB in and YUV out.

lib/src/editorhelper.cpp
```diff
@@ -234,4 +235,255 @@ std::unique_ptr<uhdr_raw_image_ext_t> apply_resize(uhdr_raw_image_t* src, int ds
...
   return std::move(dst);
 }
+
+const float BT601RGBtoYUVMatrix[5] = {0.299, 0.587, 0.114, 1.772, 1.402};
+
+const float BT709RGBtoYUVMatrix[5] = {0.2126, 0.7152, 0.0722, 1.8556, 1.5748};
+
+const float BT2020RGBtoYUVMatrix[5] = {0.2627, 0.6780, 0.0593, 1.8814, 1.4746};
+
+std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_t* src) {
```
DichenZhang1, line 245: Generally they look good to me. A few suggestions:

(1) I don't think the editorhelper is the best place for color conversion, ideally we should have a separate .h/.cpp file to keep them. (Or for now I think keep them in gainmapmath.cpp is a better choice)

(2) I think there are a lot of common code here and in gainmapmapth.h / .cpp, especially the parameters eg `BT601RGBtoYUVMatrix`. We'd better keep those parameters just in one place.

lib/src/editorhelper.cpp
ram-mohan: uhdr_raw_image_t contains a field for range. This value is configured to limited range. For yuv420p, this is configured to full range.

lib/src/editorhelper.cpp
ram-mohan: For yuv420 and p010 formats, the input is copied from source buffer to a buffer internal to the uhdr library. uhdr_enc_set_raw_image() receives a raw buffer to encode. Once the call returns it is unclear if the application can free the memory or keep it alive through the encode cycle. So currently we are doing a copy to an internal buffer and once the call returns the input can be freed or reused for other purpose. This copy code is moved from ultrahdr_api.cpp to editor_helper.cpp

lib/src/editorhelper.cpp
ram-mohan: Done

lib/src/gainmapmath.cpp
```diff
@@ -739,4 +739,260 @@ uint64_t colorToRgbaF16(Color e_gamma) {
...
+
+        r2 /= 1023.0f;
+        g2 /= 1023.0f;
+        b2 /= 1023.0f;
+
+        r3 /= 1023.0f;
+        g3 /= 1023.0f;
+        b3 /= 1023.0f;
+
+        float y0 = kYR * r0 + kYG * g0 + kYB * b0;
```
DichenZhang1, line 815: Thank you for making this change. I think there is already color transfer methods in gainmapmath.cpp eg. `bt2100YuvToRgb` `bt2100RgbToYuv`, could you re-use them? Thanks!

ram-mohan: Done.

lib/include/ultrahdr/gainmapmath.h
```diff
@@ -18,7 +18,10 @@
 #define ULTRAHDR_GAINMAPMATH_H
 
 #include <cmath>
+#include <cstring>
```
DichenZhang1, line 21: Just out of curious why we need this?

lib/src/gainmapmath.cpp
```diff
@@ -739,4 +739,183 @@ uint64_t colorToRgbaF16(Color e_gamma) {
...
+        pixel[0].v = (pixel[0].v + pixel[1].v + pixel[2].v + pixel[3].v) / 4;
+
+        pixel[0].u = (pixel[0].u * 896.0f) + 512.0f + 0.5f;
+        pixel[0].v = (pixel[0].v * 896.0f) + 512.0f + 0.5f;
+
+        pixel[0].u = CLIP3(pixel[0].u, 64.0f, 960.0f);
+        pixel[0].v = CLIP3(pixel[0].v, 64.0f, 960.0f);
+
+        uData[dst->stride[UHDR_PLANE_UV] * (i / 2) + j] = uint16_t(pixel[0].u) << 6;
+        vData[dst->stride[UHDR_PLANE_UV] * (i / 2) + j] = uint16_t(pixel[0].v) << 6;
```
DichenZhang1, line 812: Could you double check if this is right? I assume this should be `j + 1`

DichenZhang1: My mistake. I saw `vData` is defined in line 767

DichenZhang1 (approved):

DichenZhang1:
LGTM, one (maybe) redundant include

lib/include/ultrahdr/gainmapmath.h
harishdm: Dichen,
gainmapmath.cpp has memcpy which needs this.
It looks like all the headers required for cpp are included in this header and only one header is included in gainmapmath.cpp.

lib/src/editorhelper.cpp
```diff
@@ -19,6 +19,7 @@
 #include <math.h>
 
 #include "ultrahdr/editorhelper.h"
+#include "ultrahdr/gainmapmath.h"
```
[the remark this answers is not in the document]
harishdm: Removed.

DichenZhang1 (approved):

PR #111, opened 2024-04-17
refactor editor helper for adding arch specific impl

ram-mohan:
Test: ./ultrahdr_unit_test

Change-Id: I293726f4bf765987a77db1d06560f864bacb4aa9

### DIFF
--- a/lib/src/editorhelper.cpp
+++ b/lib/src/editorhelper.cpp
@@ -16,7 +16,7 @@
 
 #include <cstring>
 #include <cstdint>
-#include <math.h>
+#include <cmath>
 
 #include "ultrahdr/editorhelper.h"
 
@@ -96,14 +96,14 @@ double bicubic_interpolate(double p0, double p1, double p2, double p3, double x)
 template <typename T>
 void resize_buffer(T* src_buffer, T* dst_buffer, int src_w, int src_h, int dst_w, int dst_h,
                    int src_stride, int dst_stride, uhdr_img_fmt_t img_fmt, size_t plane) {
-  double scale_x = (double) src_w / dst_w;
-  double scale_y = (double) src_h / dst_h;
+  double scale_x = (double)src_w / dst_w;
+  double scale_y = (double)src_h / dst_h;
   for (int y = 0; y < dst_h; y++) {
     for (int x = 0; x < dst_w; x++) {
       double ori_x = x * scale_x;
       double ori_y = y * scale_y;
-      int p0_x = (int) floor(ori_x);
-      int p0_y = (int) floor(ori_y);
+      int p0_x = (int)floor(ori_x);
+      int p0_y = (int)floor(ori_y);
       int p1_x = p0_x + 1;
       int p1_y = p0_y;
       int p2_x = p0_x;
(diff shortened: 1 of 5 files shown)


lib/src/editorhelper.cpp
```diff
@@ -131,92 +146,207 @@ void resize_buffer(T* src_buffer, T* dst_buffer, int src_w, int src_h, int dst_w
...
     uint16_t* dst_buffer = static_cast<uint16_t*>(dst->planes[UHDR_PLANE_Y]);
-    rotate_buffer_clockwise(src_buffer, dst_buffer, src->w, src->h, src->stride[UHDR_PLANE_Y],
-                            dst->stride[UHDR_PLANE_Y], degree);
+    rotate_uint16_t(src_buffer, dst_buffer, src->w, src->h, -1, -1, src->stride[UHDR_PLANE_Y],
+                    dst->stride[UHDR_PLANE_Y]);
     uint32_t* src_uv_buffer = static_cast<uint32_t*>(src->planes[UHDR_PLANE_UV]);
     uint32_t* dst_uv_buffer = static_cast<uint32_t*>(dst->planes[UHDR_PLANE_UV]);
-    rotate_buffer_clockwise(src_uv_buffer, dst_uv_buffer, src->w / 2, src->h / 2,
-                            src->stride[UHDR_PLANE_UV] / 2, dst->stride[UHDR_PLANE_UV] / 2, degree);
+    rotate_uint32_t(src_uv_buffer, dst_uv_buffer, src->w / 2, src->h / 2, -1, -1,
```
DichenZhang1, line 265: This may be wrong, could you double check? Can UV interlaced be rotated the same way as non-interlaced channel?

i.e. if input is

```
u1 v1 u2 v2
u3 v3 u4 v4
```

then after 90 degree rotation it becomes:

```
u3 u1
v3 v1
u4 u2
v4 v2
```

I think the correct way is width should not be divided by 2, and sample every other pixel for U/V channel

lib/src/editorhelper.cpp
```diff
@@ -131,92 +146,207 @@ void resize_buffer(T* src_buffer, T* dst_buffer, int src_w, int src_h, int dst_w
...
+  m_mirror_horizontal_uint64_t = mirror_buffer_horizontal<uint64_t>;
+
+  m_mirror_vertical_uint8_t = mirror_buffer_vertical<uint8_t>;
+  m_mirror_vertical_uint16_t = mirror_buffer_vertical<uint16_t>;
+  m_mirror_vertical_uint32_t = mirror_buffer_vertical<uint32_t>;
+  m_mirror_vertical_uint64_t = mirror_buffer_vertical<uint64_t>;
+}
+
+uhdr_rotate_effect::uhdr_rotate_effect(int degree) : m_degree{degree} {
+  m_rotate_90_uint8_t = rotate_buffer_clockwise_90<uint8_t>;
```
DichenZhang1, line 202: I prefer making the degree a separate input parameter instead of a separate method. Reason is that we may support other degrees that are not multiple of 90 in the future after we bring OpenGL into the library.

lib/src/editorhelper.cpp
ram-mohan: Dichen, i am treating U and V as one unit and operating on the entire unit. Data type is uint32_t. So both u and v are rotated together

DichenZhang1: Ah OK. Sorry my mistake

ram-mohan: After rotation, 
u3 v3 u1 v1
u4 v4 u2 v2

lib/src/editorhelper.cpp
```diff
@@ -247,42 +377,44 @@ void apply_crop(uhdr_raw_image_t* src, int left, int top, int wd, int ht) {
...
     uint16_t* dst_buffer = static_cast<uint16_t*>(dst->planes[UHDR_PLANE_Y]);
-    resize_buffer(src_buffer, dst_buffer, src->w, src->h, dst->w, dst->h, src->stride[UHDR_PLANE_Y],
-                  dst->stride[UHDR_PLANE_Y]);
+    desc->m_resize_uint16_t(src_buffer, dst_buffer, src->w, src->h, dst->w, dst->h,
+                            src->stride[UHDR_PLANE_Y], dst->stride[UHDR_PLANE_Y]);
     uint32_t* src_uv_buffer = static_cast<uint32_t*>(src->planes[UHDR_PLANE_UV]);
     uint32_t* dst_uv_buffer = static_cast<uint32_t*>(dst->planes[UHDR_PLANE_UV]);
-    resize_buffer(src_uv_buffer, dst_uv_buffer, src->w / 4, src->h / 2, dst->w / 4, dst->h / 2,
-                  src->stride[UHDR_PLANE_UV] / 2, dst->stride[UHDR_PLANE_UV] / 2);
+    desc->m_resize_uint32_t(src_uv_buffer, dst_uv_buffer, src->w / 4, src->h / 2, dst->w / 4,
```
DichenZhang1, line 392: Should source and dest width be divided by 2 instead of 4? Could you double check?

lib/src/editorhelper.cpp
ram-mohan: Done

lib/src/editorhelper.cpp
ram-mohan: You are right dichen, this looks like mistake. I have fixed it.

DichenZhang1 (approved):

PR #114, opened 2024-04-19
cmake: Add support to cross compile to 32bit from 64bit machines

harishdm:
oss-fuzz cross compiles from x86_64 machines to 32 bit target by passing -m32 in CFLAGS. 
It doesn't set CMAKE_SYSTEM_PROCESSOR.


harishdm:
cc: @ram-mohan

ram-mohan (approved):

DichenZhang1 (approved):

PR #117, opened 2024-04-22
Reorder CMake architecture detection

cmacdonald-arm:
Fix bug to allow cross compiling using clang for targeting arm64. Previously when cross compiling using clang with the aarch64-linux-gnu triple, the build would fail due to the -march=armv7-a option being unsupported with the aarch64-linux-gnu triple.

Change-Id: Ib854997886c257efd7a5ad308533782081ca5900


ram-mohan (commented):
I am unable to understand why arch is initialized to arm during 64 bit compilation. Can you please let me know why this change is necessary.
Understood, the issue. Change looks good to me.

ram-mohan (approved):

DichenZhang1 (approved):

DichenZhang1:
Hi @cmacdonald-arm, thank you for your change, it looks good to me. I saw there was a recent [change](https://github.com/google/libultrahdr/commit/0f39c18cadc98cb485a38e5cbd03d7567eb8973a) from @harishdm and it may be targeting the same issue and it's already merged. Could you try if the latest library still has the same problem? Thanks!

harishdm:
@DichenZhang1 this change is still needed as arm was getting preferred over aarch64.
@cmacdonald-arm can you please rebase this change on the latest commit (as these lines changed in a recent commit)?
Thank you.

PR #124, opened 2024-04-29
Yuv color gamut conversion and transformYuv420 using Arm Neon.

cmacdonald-arm:
These 2 patches add a Arm Neon version for the YUV color gamut conversion as well as the transformYuv420 function.

### DIFF
--- a/lib/include/ultrahdr/gainmapmath.h
+++ b/lib/include/ultrahdr/gainmapmath.h
@@ -26,6 +26,10 @@
 #include "ultrahdr/ultrahdr.h"
 #include "ultrahdr/jpegr.h"
 
+#if (defined(UHDR_ENABLE_INTRINSICS) && (defined(__ARM_NEON__) || defined(__ARM_NEON)))
+#include <arm_neon.h>
+#endif
+
 #define CLIP3(x, min, max) ((x) < (min)) ? (min) : ((x) > (max)) ? (max) : (x)
 
 namespace ultrahdr {
@@ -440,6 +444,23 @@ extern const std::array<float, 9> kYuvBt2100ToBt709;
 extern const std::array<float, 9> kYuvBt2100ToBt601;
 
 Color yuvColorGamutConversion(Color e_gamma, const std::array<float, 9>& coeffs);
+
+#if (defined(UHDR_ENABLE_INTRINSICS) && (defined(__ARM_NEON__) || defined(__ARM_NEON)))
+
+extern const int16_t kYuv709To601_coeffs_neon[8];
+extern const int16_t kYuv709To2100_coeffs_neon[8];
+extern const int16_t kYuv601To709_coeffs_neon[8];
+extern const int16_t kYuv601To2100_coeffs_neon[8];
+extern const int16_t kYuv2100To709_coeffs_neon[8];
+extern const int16_t kYuv2100To601_coeffs_neon[8];
+
+/*
+ * The Y values are provided at half the width of U & V values to allow use of the widening
+ * arithmetic instructions.
+ */
+int16x8x3_t yuvConversion_neon(uint8x8_t y, int16x8_t u, int16x8_t v, int16x8_t coeffs);
+
+#endif
 
 /*
  * Performs a color gamut transformation on an entire YUV420 image.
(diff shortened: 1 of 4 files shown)


lib/src/aarch64/gainmapmath_neon.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+      vst1q_u8(y0_ptr + w * 2, y0_output);
+      vst1q_u8(y1_ptr + w * 2, y1_output);
+      vst1_u8(u_ptr + w, u_output);
+      vst1_u8(v_ptr + w, v_output);
+
+      w += 8;
+    } while (w < image->width / 2);
+    y0_ptr += image->luma_stride * 2;
+    y1_ptr += image->luma_stride * 2;
```
DichenZhang1, line 165: I think this may go out of boundary if the image height is odd. Could you double check?

ram-mohan: This change conflicts with https://github.com/google/libultrahdr/pull/126/commits
@cmacdonald-arm, few observations, 
- folder structure is slightly different. I am using lib/src/dsp/.  Is lib/src/dsp/arm/  better ? (will be similar to libvpx)
- enabled aosp build (Android.bp as well)
- option to disable all intrinsics via macro at configure level
- support all resolutions, widths may not align to 16 always, example 1080x1920.
pls let me know if you have any suggestions.

cmacdonald-arm: Hi Ram,

Personally I have no issues with the creation of a dsp/arm directory. I named the directory AArch64 to signify that all code is only 64-bit compatible. Also if some redesign is taking place to help with the integration of Neon implementations this could help greatly.

With your final point, images which are not aligned to 16 should be fine as along as the buffer containing the images is padded? Currently the ALIGNM macro is used to allow a multiple of 16 scan-lines to be passed to the jpeg encoder, is this likely to change?

ALIGNM is used in API-0: https://github.com/google/libultrahdr/blob/main/lib/src/jpegr.cpp#L225 
and API-1: https://github.com/google/libultrahdr/blob/main/lib/src/jpegr.cpp#L225

ram-mohan: aligned to 16 is an acceptable assumption. This is not likely to change. So the intrinsics should be fine. I have updated the folder structure to lib/src/dsp/arm/*.cpp

cmacdonald-arm: Hi Ram,

Would the prefered approach here be, first merge your [PR](https://github.com/google/libultrahdr/pull/126/commits), with the updates to the build system and file structure/names. Second rebase these changes and rename the aarch64 Neon files?

ram-mohan: As pull request #126 contains build changes for different platforms (android and host), may be, it can be merged first. Then this change can be rebased on top of main.
We have introduced a macro UHDR_ENABLE_INTRINSICS to enable/disable intrinsics if required. I think that also needs to be incorporated in your change. Currently the __arch64__ is the only gating option.
There is gainmapmath_neon.h, can this be merged with gainmapmath.h just like editor_helper.h. This is to avoid more header files. If not, when x86 intrinsics are introduced there will be more header files.

ram-mohan: The change looks good to me. Thank you.

jwright-arm:
@ram-mohan, @DichenZhang1 Do we need to do anything else here before this can be merged?

ram-mohan:
@jwright-arm The changes look good to me. I will check perf numbers with and with out the changes once. Nothing is required from you side. Thank you.

ram-mohan:
@cmacdonald-arm @jwright-arm I have profiled from my side. The numbers look really good. Thank you for this change. Couple of observations, with neon enabled the change is not bit exact. Is this expected. Further i am seeing an improvement of 16x for convertYuvNeon method. This seems to be on the larger side because neon module is processing 8 pixels per iteration in comparison with non-vector implementation which is doing 1 per iteration. I am unable to explain the additional 2x gain.

cmacdonald-arm:
@ram-mohan This change is expected to not to be bit exact, it contains a possible off by one errors. This is due to the usage of fixed point rather than floating point coefficients. In terms of the uplift, 16 Y values and 8 U & V values are processed per loop iteration, also with the usage of fixed point, conversions from integer to floating point and vice versa are no longer needed.

ram-mohan:
Thank you, @DichenZhang1 This can be merged as well.

DichenZhang1 (approved):

DichenZhang1:
Thank you! @cmacdonald-arm @ram-mohan

PR #132, opened 2024-05-07
Support ISO 21496-1 gain map metadata encode and decode

DichenZhang1:


lib/include/ultrahdr/gainmapmetadata.h
```diff
@@ -0,0 +1,99 @@
...
+                                          gain_map_metadata* out_gain_map_metadata);
+
+    static status_t gainmapMetadataFractionToFloat(const gain_map_metadata* from,
+                                                   ultrahdr_metadata_ptr to);
+
+    static status_t gainmapMetadataFloatToFraction(const ultrahdr_metadata_ptr from,
+                                                   gain_map_metadata* to);
+
+    void dump() const {
+      printf("GAIN MAP METADATA: \n");
```
ram-mohan, line 62: ALOGD is preferred. This helps android builds as well

lib/include/ultrahdr/gainmapmetadata.h
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
ram-mohan, line 2: should 2022 be modified to 2024?

lib/include/ultrahdr/gainmapmetadata.h
ram-mohan: ultrahdrcommon.h defines alogd to print for non aosp builds.

lib/include/ultrahdr/jpegr.h
```diff
@@ -27,7 +27,7 @@
 namespace ultrahdr {
 
 // The current JPEGR version that we encode to
-static const char* const kJpegrVersion = "1.0";
+static const char* const kJpegrVersion = kGainMapVersion;
```
ram-mohan, line 30: nit; can this change be used to correct all references to "1.0" to kJpegRVersion
appendGainMap routine has 
`  if (metadata->version.compare("1.0")) {
    ALOGE("received bad value for version: %s", metadata->version.c_str());
    return ERROR_JPEGR_BAD_METADATA;
  }`
jpegr_test.cpp has
`  good_metadata.version = "1.0";` and
`  metadata_expected.version = "1.0";`

Further if we intend to remove kJpegrVersion completely in the future (once aosp changes are made) because it seems to be an alias to kGainMapVersion.

lib/include/ultrahdr/jpegrutils.h
```diff
@@ -20,6 +20,10 @@
 #include "ultrahdr/ultrahdr.h"
 #include "ultrahdr/jpegr.h"
 
+// TODO (dichenzhang): This is old version metadata, new version can be foundin in
+// https://drive.google.com/file/d/1yUGmjGytRuBa2vpr9eM5Uu8CVhyyddjp/view?resourcekey=0-HGzFrzPQzu5FNYLRAJXQBA
```
ram-mohan, line 24: nit; found

lib/src/gainmapmath.cpp
```diff
@@ -937,4 +938,71 @@ std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_
...
+  // For max_iter >= 39 we shouldn't get here. max_iter can be set
+  // to a lower value to speed up the algorithm if needed.
+  *numerator = (uint32_t)round((float)(*denominator) * v);
+  return true;
+}
+
+bool floatToSignedFraction(float v, int32_t* numerator, uint32_t* denominator) {
+  uint32_t positive_numerator;
+  if (!floatToUnsignedFractionImpl(fabs(v), INT32_MAX, &positive_numerator, denominator)) {
+      return false;
```
ram-mohan, line 994: nit; indentation is not uniform through out the file.  some places in the file there are 4 spaces and some other 2 spaces. best to use once `clang-format --style=file lib/src/gainmapmath.cpp > out.txt`

lib/src/gainmapmetadata.cpp
```diff
@@ -0,0 +1,309 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
ram-mohan, line 2: nit; 2024

tests/gainmapmetadata_test.cpp
```diff
@@ -0,0 +1,83 @@
...
+  EXPECT_EQ(expected.minContentBoost, decodedUHdrMetadata.minContentBoost);
+  EXPECT_EQ(expected.gamma, decodedUHdrMetadata.gamma);
+  EXPECT_EQ(expected.offsetSdr, decodedUHdrMetadata.offsetSdr);
+  EXPECT_EQ(expected.offsetHdr, decodedUHdrMetadata.offsetHdr);
+  EXPECT_EQ(expected.hdrCapacityMin, decodedUHdrMetadata.hdrCapacityMin);
+  EXPECT_EQ(expected.hdrCapacityMax, decodedUHdrMetadata.hdrCapacityMax);
+
+
+//  const std::string kIsoNameSpace = "urn:iso:std:iso:ts:21496:-1";
+//  printf("[dichenzhang] sizeof(kIsoNameSpace)=%zu\n", sizeof(kIsoNameSpace));
```
ram-mohan: nit; looks debug log can be removed?

tests/gainmapmetadata_test.cpp
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
ram-mohan, line 2: 2024?

lib/include/ultrahdr/jpegr.h
ram-mohan: we can add a todo

DichenZhang1: I think we can keep this as is for now since the format is not finalized yet. I defined `kGainMapVersion` and `kJpegrVersion` and they may be different in the future. e.g. `kGainMapVersion` can be "2.0" and iso metadata will use 2.0 version, while `kJpegrVersion` can still be "1.0" and XMP uses 1.0 metadata to keep the backward compatibility.

lib/src/gainmapmetadata.cpp
```diff
@@ -0,0 +1,305 @@
...
+  to->version = kGainMapVersion;
+  to->maxContentBoost = from->gainMapMaxN[0] / from->gainMapMaxD[0];
+  to->minContentBoost = from->gainMapMinN[0] / from->gainMapMinD[0];
+  to->gamma = from->gainMapGammaN[0] / from->gainMapGammaD[0];
+
+  // BaseRenditionIsHDR is false
+  to->offsetSdr = from->baseOffsetN[0] / from->baseOffsetD[0];
+  to->offsetHdr = from->alternateOffsetN[0] / from->alternateOffsetD[0];
+  to->hdrCapacityMax = from->alternateHdrHeadroomN / from->alternateHdrHeadroomD;
+  to->hdrCapacityMin = from->baseHdrHeadroomN / from->baseHdrHeadroomD;
```
haixia-meta, line 263: These are integer divisions (both numerator and denominator are uint32_t) so they will not produce the correct float value. You need to cast at least one of the two operands into a float.

lib/src/gainmapmath.cpp
```diff
@@ -937,4 +1113,70 @@ std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_
...
+
+  // Maximum denominator: makes sure that the numerator is <= maxNumerator and the denominator
+  // is <= UINT32_MAX.
+  const uint64_t maxD = (v <= 1) ? UINT32_MAX : (uint64_t)floor(maxNumerator / v);
+
+  // Find the best approximation of v as a fraction using continued fractions, see
+  // https://en.wikipedia.org/wiki/Continued_fraction
+  *denominator = 1;
+  uint32_t previousD = 0;
+  float currentV = v - floor(v);
```
haixia-meta, line 1130: I'd suggest to change all intermediate float results to double type for better precision.

IEEE single-precision float only has 23 bits of significand which is not enough to carry out continued fraction for numerator and denominators as 32-bit integers. To make matters worse, you have a while loop with up to 39 iterations, so errors will accumulate and become unacceptably large.

PR #134, opened 2024-05-13
update default values in sample app

ram-mohan:


harishdm (approved):

examples/ultrahdr_app.cpp
```diff
@@ -1172,37 +1172,44 @@ void UltraHdrAppInput::computeYUVSdrPSNR() {
...
   fprintf(stderr, "\n## encoder options : \n");
   fprintf(stderr, "    -p    raw 10 bit input resource \n");
   fprintf(stderr,
           "    -y    raw 8 bit input resource, optional. \n"
           "          if not provided tonemapping happens internally. \n");
   fprintf(stderr,
-          "    -a    raw 10 bit input resource color format. Required if raw hdr intent is "
-          "provided. [0:p010, 5:rgb1010102] \n");
+          "    -a    raw 10 bit input resource color format, optional. [0:p010, 5:rgb1010102 "
+          "(default)] \n");
```
DichenZhang1, line 1183: Could you double check / confirm the default color format is RGB instead of YUV? And besides this should be `rgba1010102`, seems like letter 'a' is missing

ram-mohan: Dichen, I have addressed rgb1010102 to rgba1010102. I have checked and the default color format is rgb.

DichenZhang1 (approved):

PR #138, opened 2024-05-14
Add vector casts to editorhelper_neon.cpp

cmacdonald-arm:
Fix bug to allow compiling with GCC 12.3.0, this was caused due to -flax-vector-conversions being disabled by default in GCC 12.3.0

Change-Id: Iabc9582ab3b59d3cf43e1a92db71338247838244


ram-mohan (approved):

ram-mohan (approved):
Thank you

harishdm (approved):

DichenZhang1 (approved):

PR #146, opened 2024-05-21
Corrected comment referencing incorrect function name

MajorMurphy:
Referencing uhdr_decode() instead of uhdr_decompress()


DichenZhang1 (approved):

DichenZhang1:
Thank you for the fix!

Issue #152, opened 2024-05-29
Google Pixel8 images fail to open


MajorMurphy:
Images generated with the Pixel8 Pro camera fail to open here:  
https://github.com/google/libultrahdr/blob/94fbd77105cf5dc0e73b85060b2a065576bbf7a6/lib/src/jpegr.cpp#L1143
Image: 4080x3072 (ratio 1.318125)
Gain Map: 540x406 (ratio 1.33004...)

![PXL_20240318_183908407](https://github.com/google/libultrahdr/assets/1276140/dcf293c0-2105-4511-a8c8-aa7491fad242)

I'll to share my solution as a pull request.

DichenZhang1:
Hi @MajorMurphy , thank you for raising this issue and providing the fix. Could you send me the image attached? (cannot be opened or downloaded on my side). Thank you!

DichenZhang1:
The issue should be fixed by https://github.com/google/libultrahdr/pull/154, could you give it a try and let me know the result? Thank you!

DichenZhang1:
@MajorMurphy thanks for the reply. Never mind I know this issue and it should already been fixed by my recent PRs.

Issue #91, opened 2024-04-01
I am getting error -20008 when trying to decode image from ultrahdr


mcDandy:
Image taken on telephoto or ultrawide lens of pixel 6 pro does give this error. normal lens works fine.

![PXL_20240401_173044293 RAW-01 COVER](https://github.com/google/libultrahdr/assets/18588943/f567d574-d50f-45c1-bd7f-946d4af1eb28)

I checked the image and everything appears to be correct. Image has 4:2:0 chroma, google photos tell that it is UltraHDR.

ram-mohan:
Thank you for reporting this. We know why this is happening. Recently we have made few changes that is causing this problem. We are working on addressing the same. Can you please share a link to access this image

mcDandy:
https://github.com/google/libultrahdr/assets/18588943/f567d574-d50f-45c1-bd7f-946d4af1eb28

Github does not change uploaded image in any way. Will this suffice?

ram-mohan:
extremely sorry for this additional work, the earlier shared image was sufficient. sorry again. will publish changes that would resolve this problem soon. thank you.

mcDandy:
Dw. I posted the bug report under thé haze of sleepiness with replay after
5,5 hours of sleep...

Dne út 2. 4. 2024 7:16 uživatel ram ***@***.***> napsal:

MajorMurphy:
@mcDandy, my solution #153 also fixes compatibility with your sample image file

PR #154, opened 2024-05-30
Supports non-integer gainmap scale factor

DichenZhang1:


MajorMurphy:
Confirmed fixes #152 and #91 now

harishdm (approved):

PR #156, opened 2024-05-30
Fixed issue checking non-integer gainmap scale factor

MajorMurphy:
Pull request #154 fixes issue #152 but not issue #91 

This pull request fixes the sample files from both issues.


MajorMurphy:
Fixed instead in 58f1185

Issue #157, opened 2024-06-01
Encode Scenario 4 : Missing Parameters


pmonck:
Using the example command:
ultrahdr_app -m 0 -i cosmat_1920x1080_420_8bit.jpg -g cosmat_1920x1080_420_8bit.jpg -f metadata.cfg

results in:
did not receive valid image width for encoding. width :  0

It looks like the example should be modified to include width (-w) and height (-h) parameters as in the other encode scenarios.

DichenZhang1:
Hi @pmonck,

Thank you for raising this issue! The width/height should be read from in input JPEG images instead of reading from the configuration. However, this indeed a bug in the app, and it should be fixed by https://github.com/google/libultrahdr/pull/159

Thank you!

DichenZhang1:
Close for now. Please re-open if there is further issue. Thank you!

pmonck:
This issue seems to have reverted in the latest commit.

ram-mohan:
looks like these changes were never merged in main. They were accidentally merged to a dev branch and the issue got closed.

pmonck:
Thanks for looking into this. How should the issue be resolved?

PR #159, opened 2024-06-03
Update ultrahdr_app.cpp

DichenZhang1:
For encoding scenario 4 the width/height should be read from the input jpeg image instead of from users' input

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -1381,12 +1381,12 @@ int main(int argc, char* argv[]) {
     }
   }
   if (mode == 0) {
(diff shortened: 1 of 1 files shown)


DichenZhang1:
This should fix https://github.com/google/libultrahdr/issues/157

pmonck:
https://github.com/google/libultrahdr/issues/157 seems to have reappeared in the latest version

DichenZhang1:
Not sure why the old PR didn't merge into main. I've merged a new PR https://github.com/google/libultrahdr/pull/212 and now the fix should be in main.

PR #174, opened 2024-06-15
API: change a parameter type from bool to int

dbermond:
Commit 073d5c7580072c246e3d6f34eee2a9200b0d0d4e added the function `uhdr_enc_set_using_multi_channel_gainmap()` to the public API, which used a `bool` type in one of its parameters.

[But `bool` is a keyword only since C23](https://en.cppreference.com/w/c/keyword/bool). Before C23, `bool` is defined in the header `<stdbool.h>`, and this header is missing from the mentioned commit. Client applications that are not compiling in C23 (or later) mode will encounter an error when including the `<ultrahdr_api.h>` header, saying that `bool` is an unknown type.

For example, compiling ImageMagick with ultrahdr support gives the following error:

```
In file included from coders/uhdr.c:47:
/usr/include/ultrahdr_api.h:320:72: error: unknown type name 'bool'
  320 |                                                                        bool use_multi_channel_gainmap);
      |                                                                        ^~~~
/usr/include/ultrahdr_api.h:1:1: note: 'bool' is defined in header '<stdbool.h>'; this is probably fixable by adding '#include <stdbool.h>'
  +++ |+#include <stdbool.h>
    1 | /*
```

This can be fixed either by including the `<stdbool.h>` header after a positive check for pre-C23 mode, or by changing the argument type to  `int` like is made with other function declarations.
    
When compiling in C++ mode, this is not needed as `bool` is a keyword in C++.

### DIFF
--- a/lib/src/ultrahdr_api.cpp
+++ b/lib/src/ultrahdr_api.cpp
@@ -496,7 +496,7 @@ void uhdr_release_encoder(uhdr_codec_private_t* enc) {
 }
 
 UHDR_EXTERN uhdr_error_info_t uhdr_enc_set_using_multi_channel_gainmap(uhdr_codec_private_t* enc,
-                                                                       bool use_multi_channel_gainmap) {
+                                                                       int use_multi_channel_gainmap) {
   uhdr_error_info_t status = g_no_error;
   uhdr_encoder_private* handle = dynamic_cast<uhdr_encoder_private*>(enc);
   if (handle == nullptr) {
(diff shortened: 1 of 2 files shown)


ram-mohan:
@dbermond Thank you for pointing out this issue. Could you please modify the argument type to int instead of including stdbool.h and force push, So that this remaining in sync with other declarations.
for instance, int is_uhdr_image(...), this was also intended to be bool but we preferred int to avoid inclusions.

dbermond:
@ram-mohan thank you for the reply. As requested, I've changed the the parameter type in the public API from `bool` to `int`. Also updated the title and original post to reflect this.

The private/internal corresponding variables are left unchanged, so the compiler makes implicit conversions with them. Please let me know if you want me to also change these ones to avoid implicit conversions.

ram-mohan:
The change looks good to me.

ram-mohan (approved):

PR #176, opened 2024-06-17
Update library to use definitions of ultrahdr_api everywhere

ram-mohan:
This is a major change but maintains bitexactness with previous commit. This change unifies legacy structure definitions with definitions of ultrahdr_api.h. This helps for better extensibility for new features and avoid redundancy.

Legacy structures are moved to ultrahdr.h. These are deprecated and only retained for backward compatibility.

Briefly,
- unify ultrahdr_color_gamut with uhdr_color_gamut_t
- unify ultrahdr_transfer_function with uhdr_color_transfer_t
- unify ultrahdr_metadata_struct with uhdr_gainmap_metadata_t
- unify jpegr_uncompressed_struct with uhdr_raw_image_t
- unify jpegr_compressed_struct with uhdr_compressed_image_t
- unify jpegr_exif_struct with uhdr_mem_block_t
- unify status_t with uhdr_error_info_t
- Deprecate ultrahdr_output_format
- Added methods to Jpeg*Helper to simplify data translation between helper
  and its users

Bug fix,
- For images with multi channel gainmap, decoded gainmap is not copied
  completely for getter functions. This is fixed.
- Add support for tonemapping linear transfer inputs
- Fixes oss-fuzz: 69287

Test: ./ultrahdr_unit_test
Test: ./ultrahdr_dec_fuzzer
Test: ./ultrahdr_enc_fuzzer

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -253,11 +253,9 @@ class UltraHdrAppInput {
                    uhdr_color_gamut_t sdrCg = UHDR_CG_BT_709,
                    uhdr_color_transfer_t hdrTf = UHDR_CT_HLG, int quality = 95,
                    uhdr_color_transfer_t oTf = UHDR_CT_HLG,
-                   uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102,
-                   bool use_full_range_color_hdr = false,
-                   int gainmap_scale_factor = 4,
-                   bool use_multi_channel_gainmap = false,
-                   int gainmap_compression_quality = 85)
+                   uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102, bool isHdrCrFull = false,
+                   int gainmapScaleFactor = 4, int gainmapQuality = 85,
+                   bool enableMultiChannelGainMap = false)
       : mHdrIntentRawFile(hdrIntentRawFile),
         mSdrIntentRawFile(sdrIntentRawFile),
         mSdrIntentCompressedFile(sdrIntentCompressedFile),
@@ -275,10 +273,10 @@ class UltraHdrAppInput {
         mQuality(quality),
         mOTf(oTf),
         mOfmt(oFmt),
-        mFullRange(use_full_range_color_hdr),
-        mMapDimensionScaleFactor(gainmap_scale_factor),
-        mMapCompressQuality(gainmap_compression_quality),
-        mUseMultiChannelGainMap(use_multi_channel_gainmap),
+        mFullRange(isHdrCrFull),
+        mMapDimensionScaleFactor(gainmapScaleFactor),
+        mMapCompressQuality(gainmapQuality),
+        mUseMultiChannelGainMap(enableMultiChannelGainMap),
         mMode(0){};
 
   UltraHdrAppInput(const char* uhdrFile, const char* outputFile,
@@ -298,6 +296,10 @@ class UltraHdrAppInput {
         mQuality(95),
         mOTf(oTf),
         mOfmt(oFmt),
+        mFullRange(UHDR_CR_UNSPECIFIED),
+        mMapDimensionScaleFactor(4),
+        mMapCompressQuality(85),
+        mUseMultiChannelGainMap(false),
         mMode(1){};
 
   ~UltraHdrAppInput() {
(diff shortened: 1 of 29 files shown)


lib/include/ultrahdr/ultrahdr.h
```diff
@@ -111,6 +116,69 @@ struct ultrahdr_metadata_struct {
...
+  // Pixel format.
+  uhdr_img_fmt_t pixelFormat = UHDR_IMG_FMT_UNSPECIFIED;
+  // Color range.
+  uhdr_color_range_t colorRange = UHDR_CR_UNSPECIFIED;
+};
+
+/*
+ * Holds information for compressed image or gain map.
+ */
+struct jpegr_compressed_struct {
```
DichenZhang1, line 158: Should this also changed to "uhdr_*"

DichenZhang1:
Thank you for the change, but we may need to hold on for a little bit. jpegr.h layer is currently used by Android, and this change will also need change in Android and it's out of scope now. We will hold this change for a while and land it with v2 version.

DichenZhang1:
I would recommend separate this change land the bug fixing part now, and hold the API change part. Let me know what you think. Thank you!

lib/include/ultrahdr/ultrahdr.h
ram-mohan: This is not required. I have kept the legacy type names as is with out making any changes to maintain compatibility with android. Earlier, the uhdr_compressed_image_t is converted to jpegr_compressed_struct and sent to decodeJPEGR. Now the reverse happens. Through out the library this change is now using uhdr_* structs and the legacy structures are mapped to uhdr_* structs.

For example,
encode API-0
  uhdr_error_info_t encodeJPEGR(uhdr_raw_image_t* hdr_intent, uhdr_compressed_image_t* dest,
                                int quality, uhdr_mem_block_t* exif) { ... } 
is the current implementation.

 // legacy implementation of encode API - 0
  status_t encodeJPEGR(jr_uncompressed_ptr p010_image_ptr, ultrahdr_transfer_function hdr_tf,
                       jr_compressed_ptr dest, int quality, jr_exif_ptr exif) { ... } is the alias of the above. 

This function maps its arguments to uhdr_* and calls current implementation.

To summarize, ultrahdr.h contains all the legacy structures and are marked for deprecation.

ram-mohan:
> Thank you for the change, but we may need to hold on for a little bit. jpegr.h layer is currently used by Android, and this change will also need change in Android and it's out of scope now. We will hold this change for a while and land it with v2 version.

This change does not effect android builds. The android builds works normally.

DichenZhang1:
OK Thanks, I'll take a further look

DichenZhang1:
> > Thank you for the change, but we may need to hold on for a little bit. jpegr.h layer is currently used by Android, and this change will also need change in Android and it's out of scope now. We will hold this change for a while and land it with v2 version.
> 
> This change does not effect android builds. The android builds works normally. It seems recent sample app changes are causing build issues in android. These were corrected in this commit.
> 
> > I would recommend separate this change land the bug fixing part now, and hold the API change part. Let me know what you think. Thank you!
> 
> Some fixes work with the changes made. I will try to seperate independent one's and give a pull request.

OK then we are having two interfaces of the public APIs. I'm not against with this move, but I'm feeling this is not an urgent need. And additionally this change is fairly too large and I would still recommend separating this change and we'll merge the bug fixing part first. Let me know what you think. thank you!

ram-mohan:
> OK then we are having two interfaces of the public APIs. 

As the interface is from ultrahdr_api.h, there is only one. But if we were to include jpegr.h directly then as you have rightly pointed out, there are aliases for each api.

> I'm not against with this move, but I'm feeling this is not an urgent need. And additionally this change is fairly too large and I would still recommend separating this change and we'll merge the bug fixing part first. 

Certain fixes that are independent of this change, i have moved them out and issued pull requests. #178 #179 #180 
But some fixes where a little more coupled with this change (oss-fuzz 69287 and returning decoded gainmap). So i did not move them. Also, some parts of this change simplifies handling 444 and 422 support. 

> When I use the attached sources, I get error "encountered unknown error during decoding" (exit code 255). 
I assume this is related to the JPG, though it does not say which file decoding failed (would be nice to update the error message to be more clear there).

It also addresses this.

Some portions of this change are not at all a priority and can be marked as good to have and some other parts are helpful towards addressing known issues.

DichenZhang1:
OK my only concern of this change is about having aliases at jpegr.cpp layer, but indeed this change makes the code unified and clean, and has benefit of simplifying 444 and 422 color format. I've approved this change. Thank you for work!

DichenZhang1 (approved):

examples/ultrahdr_app.cpp
```diff
@@ -298,6 +296,10 @@ class UltraHdrAppInput {
         mQuality(95),
         mOTf(oTf),
         mOfmt(oFmt),
+        mFullRange(UHDR_CR_UNSPECIFIED),
+        mMapDimensionScaleFactor(1),
```
DichenZhang1, line 300: Same comment as I mentioned in the other CL

DichenZhang1 (approved):

examples/ultrahdr_app.cpp
ram-mohan: Done

ram-mohan: Done

DichenZhang1 (approved):

PR #179, opened 2024-06-19
Update error checks for newly advertised encoder options

ram-mohan:
- Add error checks for input params of newly added setter functions
- update documentation of sample app for newly added options
- This change also fixed aosp builds

Test: ./ultrahdr_unit_test

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -253,11 +253,9 @@ class UltraHdrAppInput {
                    uhdr_color_gamut_t sdrCg = UHDR_CG_BT_709,
                    uhdr_color_transfer_t hdrTf = UHDR_CT_HLG, int quality = 95,
                    uhdr_color_transfer_t oTf = UHDR_CT_HLG,
-                   uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102,
-                   bool use_full_range_color_hdr = false,
-                   int gainmap_scale_factor = 4,
-                   bool use_multi_channel_gainmap = false,
-                   int gainmap_compression_quality = 85)
+                   uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102, bool isHdrCrFull = false,
+                   int gainmapScaleFactor = 4, int gainmapQuality = 85,
+                   bool enableMultiChannelGainMap = false)
       : mHdrIntentRawFile(hdrIntentRawFile),
         mSdrIntentRawFile(sdrIntentRawFile),
         mSdrIntentCompressedFile(sdrIntentCompressedFile),
@@ -275,10 +273,10 @@ class UltraHdrAppInput {
         mQuality(quality),
         mOTf(oTf),
         mOfmt(oFmt),
-        mFullRange(use_full_range_color_hdr),
-        mMapDimensionScaleFactor(gainmap_scale_factor),
-        mMapCompressQuality(gainmap_compression_quality),
-        mUseMultiChannelGainMap(use_multi_channel_gainmap),
+        mFullRange(isHdrCrFull),
+        mMapDimensionScaleFactor(gainmapScaleFactor),
+        mMapCompressQuality(gainmapQuality),
+        mUseMultiChannelGainMap(enableMultiChannelGainMap),
         mMode(0){};
 
   UltraHdrAppInput(const char* uhdrFile, const char* outputFile,
@@ -298,6 +296,10 @@ class UltraHdrAppInput {
         mQuality(95),
         mOTf(oTf),
         mOfmt(oFmt),
+        mFullRange(UHDR_CR_UNSPECIFIED),
+        mMapDimensionScaleFactor(4),
+        mMapCompressQuality(85),
+        mUseMultiChannelGainMap(false),
         mMode(1){};
 
   ~UltraHdrAppInput() {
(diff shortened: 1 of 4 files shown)


examples/ultrahdr_app.cpp
```diff
@@ -631,16 +633,13 @@ bool UltraHdrAppInput::encode() {
   RET_IF_ERR(uhdr_enc_set_using_multi_channel_gainmap(handle, mUseMultiChannelGainMap))
   RET_IF_ERR(uhdr_enc_set_gainmap_scale_factor(handle, mMapDimensionScaleFactor))
 #ifdef PROFILE_ENABLE
-  const int profileCount = 10;
   Profiler profileEncode;
   profileEncode.timerStart();
-  for (auto i = 0; i < profileCount; i++) {
```
DichenZhang1, line 637: Curious why we are removing this logic?

examples/ultrahdr_app.cpp
```diff
@@ -1240,10 +1235,18 @@ static void usage(const char* name) {
...
           "    -q    quality factor to be used while encoding 8 bit image, optional. [0-100], 95 : "
-          "default.\n"
-          "          gain map image does not use this quality factor. \n"
-          "          for now gain map image quality factor is not configurable. \n");
+          "default.\n");
   fprintf(stderr, "    -e    compute psnr, optional. [0:no (default), 1:yes] \n");
+  fprintf(stderr,
+          "    -R    color range for the HDR YUV input, optional. [0:narrow-range (default), "
+          "1:full-range]. \n");
+  fprintf(stderr, "    -s    gain map scale factor, optional. [1, 2, 3, 4 (default)]. \n");
```
DichenZhang1, line 1243: Any integer should be OK, 4 is default

examples/ultrahdr_app.cpp
```diff
@@ -1240,10 +1235,18 @@ static void usage(const char* name) {
...
+  fprintf(stderr,
+          "    -Q    quality factor to be used while encoding gain map image, optional. [0-100], "
+          "85. \n");
```
DichenZhang1, line 1246: add "default"

lib/src/ultrahdr_api.cpp
```diff
@@ -495,33 +495,67 @@ void uhdr_release_encoder(uhdr_codec_private_t* enc) {
...
+
+  if (dynamic_cast<uhdr_encoder_private*>(enc) == nullptr) {
     status.error_code = UHDR_CODEC_INVALID_PARAM;
     status.has_detail = 1;
     snprintf(status.detail, sizeof status.detail, "received nullptr for uhdr codec instance");
     return status;
   }
 
+  if (gainmap_scale_factor <= 0 ||
+      gainmap_scale_factor > (int)ultrahdr::kMapDimensionScaleFactorDefault) {
```
DichenZhang1, line 537: We shouldn't have this restriction

lib/src/ultrahdr_api.cpp
```diff
@@ -495,33 +495,67 @@ void uhdr_release_encoder(uhdr_codec_private_t* enc) {
...
+    status.error_code = UHDR_CODEC_INVALID_PARAM;
+    status.has_detail = 1;
+    snprintf(status.detail, sizeof status.detail,
+             "unsupported gainmap scale factor %d, expects one of {1, 2, 3, 4}",
```
DichenZhang1, line 541: same

DichenZhang1 (commented):
Some comment

examples/ultrahdr_app.cpp
ram-mohan: the first call to uhdr_decode() decodes the bitstream. Repeated calls will return instantly because the context already contains decoded output. So it wont decode again. The context needs to be reset and used again. This concept of trying decode 10 times and then averaging wont work. So i am collecting data in the first iteration only.

examples/ultrahdr_app.cpp
ram-mohan: may be any integer is not ok, because values <= 0 are not ok. Values more than 4 can cause gain map image width and height to become zero. Because kMinWidth and kMinHeight are configured to 4 x 4. There no checks in jpegr.cpp after gainmap dimensions are computed in generateGainMap.

lib/src/ultrahdr_api.cpp
ram-mohan: explained earlier

examples/ultrahdr_app.cpp
DichenZhang1: OK I see.

I think we can leave the low-boundary to 1, and leave the up-boundary open, and pre-check / print an error message if it's too large and triggered the gain map width / height to becoming 0.

ram-mohan: Done

DichenZhang1: Or perhaps force changing it the the largest-possible value when that happens

ram-mohan: If gainmap dimensions becomes zero i am choosing min (width, height) as scale factor.

DichenZhang1: Thank you for making this change! One minor comment that I'm feeling if this happens one pixel for the gainmap is too small to contain any useful information. What about using `min(width, height) / DCTSIZE` so that if we see extreme large scaling factor from users input we can maintain the gainmap dimension to at least 16.

ram-mohan: I have made the change dichen. As we are updating the input configuration with out feedback to the client, is it better if we update it to a value that yields the best image quality (scale factor = 1). Because if the client has configured it accidentally then without feeback he has to debug the cause of any quality issue due to small gainmap image. Also DCTSIZE is jpeg specific and if heir.cpp is introduced this can change to 64 and for avir.cpp this can change to 128. Defaulting to 1, we can update ultrahdr_api.h as well with documentation. Further if we return error if map_width or map_height is zero, then it circumvents all these thoughts.

examples/ultrahdr_app.cpp
```diff
@@ -298,6 +296,10 @@ class UltraHdrAppInput {
         mQuality(95),
         mOTf(oTf),
         mOfmt(oFmt),
+        mFullRange(UHDR_CR_UNSPECIFIED),
+        mMapDimensionScaleFactor(1),
```
DichenZhang1, line 300: A minor comment: I think it's better to keep it 4 as default here, which is aligned with the original code and other places in this change.

DichenZhang1 (approved):

examples/ultrahdr_app.cpp
ram-mohan: Done

PR #181, opened 2024-06-21
Support gamma from users' input in the demo app

DichenZhang1:


examples/ultrahdr_app.cpp
```diff
@@ -372,6 +374,7 @@ class UltraHdrAppInput {
   size_t mMapDimensionScaleFactor;
   int mMapCompressQuality;
   bool mUseMultiChannelGainMap;
+  float mGamma;
```
ram-mohan, line 377: const float mGainMapGamma; 
this keeps in sync with other constructor arg declarations

lib/include/ultrahdr/jpegr.h
```diff
@@ -137,7 +136,7 @@ class JpegR {
  public:
    JpegR();
 
-   JpegR(size_t, int, bool);
+   JpegR(size_t, int, bool, float);
```
ram-mohan, line 139: all constructors can be merged with 
  JpegR(size_t mapDimensionScaleFactor = kMapDimensionScaleFactorDefault,
        int mapCompressQuality = kMapCompressQualityDefault,
        bool useMultiChannelGainMap = kUseMultiChannelGainMapDefault,
        float gainMapGamma = kGainMapGammaDefault);

lib/include/ultrahdr/jpegr.h
```diff
@@ -332,6 +331,12 @@ class JpegR {
 
    bool isUsingMultiChannelGainMap() { return this->mUseMultiChannelGainMap; }
 
+   void setGainMapGamma(float gamma) {
```
ram-mohan, line 334: are these getter setter functions required, as these are only controlled from constructor

lib/src/ultrahdr_api.cpp
```diff
@@ -525,6 +525,21 @@ UHDR_EXTERN uhdr_error_info_t uhdr_enc_set_gainmap_scale_factor(uhdr_codec_priva
...
+                                                         float gamma) {
+  uhdr_error_info_t status = g_no_error;
+  uhdr_encoder_private* handle = dynamic_cast<uhdr_encoder_private*>(enc);
+  if (handle == nullptr) {
+    status.error_code = UHDR_CODEC_INVALID_PARAM;
+    status.has_detail = 1;
+    snprintf(status.detail, sizeof status.detail, "received nullptr for uhdr codec instance");
+    return status;
+  }
+
```
ram-mohan, line 538: error checks for gamma range are required. 
check required if context has moved to running state from init (m_sailed).

lib/include/ultrahdr/jpegr.h
DichenZhang1: I think yes. User may change the configurations on the fly since we don't have a `m_sailed`-like flag at jpegr.cpp layer.

Besides, it's not convenient to set this parameters only from constructor. E.g. if user only wants to set gamma while use the default values for the other parameters, they'll need to also provide the other parameter values in the constructor.

ram-mohan:
This can be closed as these changes are already merged?

PR #186, opened 2024-06-29
Fixed compiling ARM64 target for Windows

MajorMurphy:
Fixes two issues:

-  gcc `__attribute__` keyword not supported by MSVC
- assert looking for `width` variable instead of `w`

### DIFF
--- a/lib/src/dsp/arm/gainmapmath_neon.cpp
+++ b/lib/src/dsp/arm/gainmapmath_neon.cpp
@@ -19,6 +19,12 @@
 #include <arm_neon.h>
 #include <cassert>
 
+#ifdef _MSC_VER
+#define ALIGNED(x) __declspec(align(x))
+#else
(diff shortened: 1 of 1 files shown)


harishdm (approved):

lib/src/dsp/arm/gainmapmath_neon.cpp
```diff
@@ -19,6 +19,12 @@
 #include <arm_neon.h>
 #include <cassert>
 
+#if _WIN32
```
ram-mohan, line 22: should this be 
#ifdef _MSC_VER

MajorMurphy: Changed in newest commit

harishdm (approved):

ram-mohan (approved):
thanks. looks good to me.

PR #191, opened 2024-07-07
Improve gainmap coefficients signalling accuracy

ram-mohan:
- add rounding factor to gainmap coefficients before converting them to 8 bit ints
- capture min and max content boost basing on the computed gainmap coefficients and use them during affine mapping
- use hdr offset and sdr offset to effectively signal hdr black
- add option to configure maxCLL of hdr intent

Test: ./ultrahdr_unit_test

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -463,7 +463,6 @@ ColorTransformFn getInverseOetf(uhdr_color_transfer_t transfer) {
 float getMaxDisplayMasteringLuminance(uhdr_color_transfer_t transfer) {
   switch (transfer) {
     case UHDR_CT_LINEAR:
-      // TODO: configure maxCLL correctly for linear tf
       return kHlgMaxNits;
     case UHDR_CT_HLG:
       return kHlgMaxNits;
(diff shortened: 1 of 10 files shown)


lib/src/gainmapmath.cpp
```diff
@@ -463,7 +463,6 @@ ColorTransformFn getInverseOetf(uhdr_color_transfer_t transfer) {
 float getMaxDisplayMasteringLuminance(uhdr_color_transfer_t transfer) {
   switch (transfer) {
     case UHDR_CT_LINEAR:
-      // TODO: configure maxCLL correctly for linear tf
```
DichenZhang1, line 466: I think we need to keep the "TODO"

lib/src/jpegr.cpp
```diff
@@ -458,7 +510,9 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
     return status;
   }
 
-  float hdr_white_nits = getMaxDisplayMasteringLuminance(hdr_intent->ct);
+  float hdr_white_nits = mMaxDisplayLuminanceNits == -1
```
DichenZhang1, line 513: nit: -1.0f

lib/src/jpegr.cpp
```diff
@@ -628,6 +693,67 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+  for (size_t rowStart = 0; rowStart < map_height;) {
+    size_t rowEnd = (std::min)(rowStart + rowStep, map_height);
+    jobQueue.enqueueJob(rowStart, rowEnd);
+    rowStart = rowEnd;
+  }
+  jobQueue.markQueueForEnd();
+  encodeMap();
+  std::for_each(workers.begin(), workers.end(), [](std::thread& t) { t.join(); });
+
+  gainmap_metadata->max_content_boost = exp2(max_content_boost_log2);
```
DichenZhang1, line 749: I'm afraid this may break the decoding on Android.

Could you help to encode two jpeg/r images, with and without this change and try displaying them on Android devices (or give me the images and I'll do the verification)

lib/src/jpegr.cpp
```diff
@@ -628,6 +693,67 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+                               map_width, mUseMultiChannelGainMap);*/
+
+  float min_content_boost_log2 = gainmap_min[0];
+  float max_content_boost_log2 = gainmap_max[0];
+  for (int index = 1; index < (mUseMultiChannelGainMap ? 3 : 1); index++) {
+    min_content_boost_log2 = (std::min)(gainmap_min[index], min_content_boost_log2);
+    max_content_boost_log2 = (std::max)(gainmap_max[index], max_content_boost_log2);
+  }
+
+  std::function<void()> encodeMap = [this, gainmap_data, map_width, dest, min_content_boost_log2,
```
DichenZhang1, line 706: This change has added an additional step and I'm afraid that this may affect the performance. Could you try some profiling for generateGainMap() call before and after this change?

(We were doing this before but changed it to one step as the current version at some point, reasons are performance consideration, and below)

Also note that we saw this may NOT ideal for image quality especially when there is JPEG input because the JPEG input may have encoding artifacts

lib/src/jpegr.cpp
```diff
@@ -628,6 +693,67 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+    size_t rowStart, rowEnd;
+
+    while (jobQueue.dequeueJob(rowStart, rowEnd)) {
+      if (mUseMultiChannelGainMap) {
+        for (size_t j = rowStart; j < rowEnd; j++) {
+          size_t dst_pixel_idx = j * dest->stride[UHDR_PLANE_PACKED] * 3;
+          size_t src_pixel_idx = j * map_width * 3;
+          for (size_t i = 0; i < map_width * 3; i++) {
+            reinterpret_cast<uint8_t*>(dest->planes[UHDR_PLANE_PACKED])[dst_pixel_idx + i] =
+                affineMapGain(gainmap_data[src_pixel_idx + i], min_content_boost_log2,
```
DichenZhang1, line 717: I think we missed `gamma` here?

lib/src/gainmapmath.cpp
ram-mohan: as maxCLL is made configurable via ultrahdr api, i have remove this comment dichen

lib/src/jpegr.cpp
ram-mohan: Done

lib/src/jpegr.cpp
```diff
@@ -659,6 +724,71 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+                              max_content_boost_log2, kGainMapGammaDefault);
```
DichenZhang1, line 753: I think we need to use `mGainMapGammaDefault` here.

lib/src/jpegr.cpp
```diff
@@ -659,6 +724,71 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+          }
+        }
+      } else {
+        for (size_t j = rowStart; j < rowEnd; j++) {
+          size_t dst_pixel_idx = j * dest->stride[UHDR_PLANE_Y];
+          size_t src_pixel_idx = j * map_width;
+          for (size_t i = 0; i < map_width; i++) {
+            reinterpret_cast<uint8_t*>(dest->planes[UHDR_PLANE_Y])[dst_pixel_idx + i] =
+                affineMapGain(gainmap_data[src_pixel_idx + i], min_content_boost_log2,
+                              max_content_boost_log2, kGainMapGammaDefault);
```
DichenZhang1, line 763: same

lib/src/jpegr.cpp
```diff
@@ -659,6 +724,71 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
...
+    jobQueue.enqueueJob(rowStart, rowEnd);
+    rowStart = rowEnd;
+  }
+  jobQueue.markQueueForEnd();
+  encodeMap();
+  std::for_each(workers.begin(), workers.end(), [](std::thread& t) { t.join(); });
+
+  gainmap_metadata->max_content_boost = exp2(max_content_boost_log2);
+  gainmap_metadata->min_content_boost = exp2(min_content_boost_log2);
+  gainmap_metadata->gamma = kGainMapGammaDefault;
```
DichenZhang1, line 786: same

PR #197, opened 2024-07-11
Enable gainmap tests

ram-mohan:


harishdm (approved):

ram-mohan:
This is actually a benchmarking test. It was initially written in jpegrtest.cpp. Later we have introduced benchmark tests. However this test was never relocated there. May be we can move that there at a later time.

DichenZhang1:
Yes I agree. I can merge this CL now and move it later (not urgent). Thank you!

PR #205, opened 2024-07-15
Add option to enable compiling with -Werror

georges-arm:
The libultrahdr codebase itself compiles cleanly with `-Werror` however third-party dependent libraries do not. Since there is no way of specifying options for only libultrahdr-related code itself, add a new `UHDR_ENABLE_WERROR` option to enable passing `-Werror` for just the core libraries and dependent targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ option_if_not_defined(UHDR_ENABLE_LOGS "Build with verbose logging " FALSE)
 option_if_not_defined(UHDR_ENABLE_INSTALL "Add install and uninstall targets for libuhdr package" TRUE)
 option_if_not_defined(UHDR_ENABLE_INTRINSICS "Build with intrinsics " TRUE)
 option_if_not_defined(UHDR_ENABLE_OPENGL "Build with gpu acceleration " FALSE)
+option_if_not_defined(UHDR_ENABLE_WERROR "Build with -Werror" FALSE)
 
 # pre-requisites
 if(UHDR_BUILD_TESTS AND EMSCRIPTEN)
@@ -187,6 +188,7 @@ if(UHDR_BUILD_FUZZERS)
   add_compile_options(-fsanitize=fuzzer-no-link)
 endif()
 
+set(UHDR_WERROR_FLAGS "")
 if(MSVC)
   add_definitions(-D_CRT_SECURE_NO_WARNINGS)
   # Disable specific warnings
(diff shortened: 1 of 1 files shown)


ram-mohan:
We tried adding -Werror before and as there were few third party libraries producing warnings, we deferred it. We intend to address these warnings first and keep -werror for all targets. With this in mind, may be this change is not required?

ram-mohan:
@georges-arm All the warnings in libuhdr sources are resolved. May be we can enable this flag for  library (not third party dependencies) by default

georges-arm:
> @georges-arm All the warnings in libuhdr sources are resolved. May be we can enable this flag for library (not third party dependencies) by default

Hi Ram,

Thanks for the update. I'd be cautious about enabling `-Werror` by default since it's not possible to know if future compiler releases may enable new warnings that would then cause the build to fail for users even though the library has not changed.

Traditionally the approach I have seen taken in other projects is to have `-Werror` enabled by a flag and then have this flag enabled in CI and pre-commit to catch warnings on the current compilers without breaking the project for external users who may be using newer or less well-tested compilers which may emit different warnings.

What do you think?

Thanks,
George

ram-mohan:
> > @georges-arm All the warnings in libuhdr sources are resolved. May be we can enable this flag for library (not third party dependencies) by default
> 
> Hi Ram,
> 
> Thanks for the update. I'd be cautious about enabling `-Werror` by default since it's not possible to know if future compiler releases may enable new warnings that would then cause the build to fail for users even though the library has not changed.
> 
> Traditionally the approach I have seen taken in other projects is to have `-Werror` enabled by a flag and then have this flag enabled in CI and pre-commit to catch warnings on the current compilers without breaking the project for external users who may be using newer or less well-tested compilers which may emit different warnings.
> 
> What do you think?
> 
> Thanks, George

Sounds good to me. I think your approach is the best way forward.

georges-arm:
I've rebased and updated the commit to include adding a line to the `docs/building.md` file with the new option. I've also taken the opportunity to reorder the recently-added `UHDR_BUILD_JAVA` option to be adjacent to the other `UHDR_BUILD_*` options.

ram-mohan:
changes look good to me. Should the CI workflows in .github also be modified?

ram-mohan:
Thank you very much for the CL. every thing looks good to me.

ram-mohan (approved):

harishdm (approved):

DichenZhang1 (approved):

PR #215, opened 2024-07-20
Allow for larger excursions of min, max content boost

ram-mohan:
Current implementation clips min/max content boost to a smaller range. Allow for larger excursions for better hdr intent representation.

Also, fixed round factor before encoding gainmap coefficient

Test: ./ultrahdr_unit_test

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -244,17 +244,20 @@ static bool writeFile(const char* filename, uhdr_raw_image_t* img) {
 
 class UltraHdrAppInput {
  public:
-  UltraHdrAppInput(
-      const char* hdrIntentRawFile, const char* sdrIntentRawFile,
-      const char* sdrIntentCompressedFile, const char* gainmapCompressedFile,
-      const char* gainmapMetadataCfgFile, const char* exifFile, const char* outputFile,
-      size_t width, size_t height, uhdr_img_fmt_t hdrCf = UHDR_IMG_FMT_32bppRGBA1010102,
-      uhdr_img_fmt_t sdrCf = UHDR_IMG_FMT_32bppRGBA8888,
-      uhdr_color_gamut_t hdrCg = UHDR_CG_DISPLAY_P3, uhdr_color_gamut_t sdrCg = UHDR_CG_BT_709,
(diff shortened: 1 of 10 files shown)


DichenZhang1:
We may not be able to merge this since this may introduce other concerns (e.g. performance drag, introducing error if the input is compressed due to compression artifacts). Besides, existing components e.g. Skia is using the old logic, and moving to this logic may lead to changes outside this library. We can hold this for now.

DichenZhang1:
@ram-mohan @harishdm we will take this change, and some modifications need to be done.

We can introduce an encode mode parameter, "SPEED" mode and "QUALITY" mode, the former to be default. Under "QUALITY" mode we do this. This can happen in API-1, 2, 3 (no API-0 because our internal tone mapping already considered this)

I suggest in this change we start we a macro definition of the encode mode parameter, and making it configurable through the command line app in a follow-up change.

Thank you for your help!

DichenZhang1:
Besides, we may also try out https://github.com/google/libultrahdr/pull/236, which has similar effects for API-1 in my local tests but it's much easier.

ram-mohan:
> @ram-mohan @harishdm we will take this change, and some modifications need to be done.
> 
> We can introduce an encode mode parameter, "SPEED" mode and "QUALITY" mode, the former to be default. Under "QUALITY" mode we do this. This can happen in API-1, 2, 3 (no API-0 because our internal tone mapping already considered this)
> 
> I suggest in this change we start with a macro definition of the encode mode parameter, and making it configurable through the command line app in a follow-up change.
> 
> Thank you for your help!

Dichen, I have made the necessary changes and pushed a patchset.

ram-mohan:
> Besides, we may also try out #236, which has similar effects for API-1 in my local tests but it's much easier.

This should improve quality but solution is not universal. Even though the lower limit is extended from 0 to -6.64, to some extent it is clamped and the upper limit is still clamped. Range of content boost is indicative of dynamic range of recovery of hdr intent from sdr intent. Clamping to a predefined value can work in most cases, but there can be scenarios where this can cause limitations.

DichenZhang1:
Thank you for the latest change. My point is if #236 can give us similar quality improvement as this one, we can consider taking #236. Or we may take both.

DichenZhang1 (approved):

PR #218, opened 2024-07-28
Correct inverse oetf functions of sRGB and PQ

ram-mohan:
Some constants used in inverse oetf functions of sRGB and PQ are not accurate. These are fixed.

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -124,11 +124,11 @@ Color srgbInvOetfLUT(Color e_gamma) {
 }
 
 float srgbOetf(float e) {
-  constexpr float kThreshold = 0.00304;
+  constexpr float kThreshold = 0.0031308;
   constexpr float kLowSlope = 12.92;
   constexpr float kHighOffset = 0.055;
   constexpr float kPowerExponent = 1.0 / 2.4;
-  if (e < kThreshold) {
+  if (e <= kThreshold) {
     return kLowSlope * e;
   }
   return (1.0 + kHighOffset) * std::pow(e, kPowerExponent) - kHighOffset;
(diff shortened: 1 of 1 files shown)


lib/src/gainmapmath.cpp
```diff
@@ -123,12 +123,13 @@ Color srgbInvOetfLUT(Color e_gamma) {
   return {{{srgbInvOetfLUT(e_gamma.r), srgbInvOetfLUT(e_gamma.g), srgbInvOetfLUT(e_gamma.b)}}};
 }
 
+// From Khronos DataFormat Spec 1.3, Section 13.3.2 - sRGB Inverse EOTF
```
DichenZhang1, line 126: The number is correct. But I suggest removing this comment because we have a comment before `srgbInvOetf`. I want to make sure the transfer and inverse transfer come from the same source. This comment may cause confusion.

lib/src/gainmapmath.cpp
```diff
@@ -289,18 +290,10 @@ float pqOetfLUT(float e) {
 
 Color pqOetfLUT(Color e) { return {{{pqOetfLUT(e.r), pqOetfLUT(e.g), pqOetfLUT(e.b)}}}; }
 
-// Derived from the inverse of the Reference PQ OETF.
-static const float kPqInvA = 128.0f, kPqInvB = 107.0f, kPqInvC = 2413.0f, kPqInvD = 2392.0f,
-                   kPqInvE = 6.2773946361f, kPqInvF = 0.0126833f;
-
+// From Khronos DataFormat Spec 1.3, Section 13.6.1 - PQ Inverse OETF
```
DichenZhang1, line 293: Same comment. Please remove this comment since we have comment for `pqOetf`, though the formula is the same

lib/src/gainmapmath.cpp
ram-mohan: Done.

DichenZhang1 (approved):

PR #222, opened 2024-07-31
Add support for GPU acceleration of applygainmap

aayushsoni111:
Test: ./ultrahdr_app -m 1 -j input_uhdr.jpg -o 1 -O 5
Test: ./ultrahdr_unit_test

Change-Id: I84111573ee7ea2e32221ef063fac0496e0cd8dbe


DichenZhang1 (approved):

lib/include/ultrahdr/jpegr.h
```diff
@@ -79,10 +79,14 @@ typedef struct jpegr_info_struct* jr_info_ptr;
 
 class JpegR {
  public:
-  JpegR(size_t mapDimensionScaleFactor = kMapDimensionScaleFactorDefault,
-        int mapCompressQuality = kMapCompressQualityDefault,
-        bool useMultiChannelGainMap = kUseMultiChannelGainMapDefault,
-        float gamma = kGainMapGammaDefault);
+  JpegR(
+#ifdef UHDR_ENABLE_GLES
```
harishdm, line 83: Ram,
This change causes build failure in Android.

When this header is included from a different project, UHDR_ENABLE_GLES will need to be explicitly defined in that project's Android.bp as well.

So this macro in this header file isn't a good idea. Need to figure out how to handle this.

PR #227, opened 2024-08-05
revisit gainmapmath constants

ram-mohan:
- luminance function for p3 gamut is using weights of DCI P3 instead of Display P3. This is corrected. Accordingly, the gamut conversions involving Display P3 are updated.
- Increased precision of some constants used in csc, oetf and eotf functions.

Test: ./ultrahdr_unit_test

### DIFF
--- a/lib/include/ultrahdr/icc.h
+++ b/lib/include/ultrahdr/icc.h
@@ -103,6 +103,10 @@ static constexpr size_t kICCTagTableEntrySize = 12;
 // bytes for a single XYZ number type (4 bytes per coordinate).
 static constexpr size_t kColorantTagSize = 20;
 
+// size should be 12; 4 bytes for type descriptor, 4 bytes reserved, one
+// byte each for primaries, transfer, matrix, range.
+static constexpr size_t kCicpTagSize = 12;
+
 static constexpr uint32_t kDisplay_Profile = SetFourByteTag('m', 'n', 't', 'r');
 static constexpr uint32_t kRGB_ColorSpace = SetFourByteTag('R', 'G', 'B', ' ');
 static constexpr uint32_t kXYZ_PCSSpace = SetFourByteTag('X', 'Y', 'Z', ' ');
@@ -149,10 +153,12 @@ static constexpr Matrix3x3 kRec2020 = {{
     {-0.00193139f, 0.0299794f, 0.797162f},
 }};
 
+static constexpr uint32_t kCICPPrimariesUnSpecified = 2;
 static constexpr uint32_t kCICPPrimariesSRGB = 1;
 static constexpr uint32_t kCICPPrimariesP3 = 12;
 static constexpr uint32_t kCICPPrimariesRec2020 = 9;
 
+static constexpr uint32_t kCICPTrfnUnSpecified = 2;
 static constexpr uint32_t kCICPTrfnSRGB = 1;
 static constexpr uint32_t kCICPTrfnLinear = 8;
 static constexpr uint32_t kCICPTrfnPQ = 16;
@@ -238,7 +244,7 @@ struct ICCHeader {
 class IccHelper {
  private:
   static constexpr uint32_t kTrcTableSize = 65;
-  static constexpr uint32_t kGridSize = 17;
+  static constexpr uint32_t kGridSize = 11;
   static constexpr size_t kNumChannels = 3;
 
   static std::shared_ptr<DataStruct> write_text_tag(const char* text);
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -1585,6 +1585,72 @@ std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_
   return dst;
 }
 
+uhdr_error_info_t convert_ycbcr_input_to_rgb(uhdr_raw_image_t* src, uhdr_raw_image_t* dst) {
+  if (dst->w != src->w || dst->h != src->h) {
+    uhdr_error_info_t status;
+    status.error_code = UHDR_CODEC_MEM_ERROR;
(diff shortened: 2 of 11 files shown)


ram-mohan:
Investigate if this change affects the entries of ICC box

lib/src/gainmapmath.cpp
```diff
@@ -139,12 +150,13 @@ Color srgbOetf(Color e) { return {{{srgbOetf(e.r), srgbOetf(e.g), srgbOetf(e.b)}
 ////////////////////////////////////////////////////////////////////////////////
 // Display-P3 transformations
 
-// See SMPTE EG 432-1, Equation 7-8.
-static const float kP3R = 0.20949f, kP3G = 0.72160f, kP3B = 0.06891f;
+// Display P3 Luminance
+// See, https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#PRIMARIES_DISPLAYP3
+static const float kP3R = 0.2289746f, kP3G = 0.6917385f, kP3B = 0.0792869f;
```
DichenZhang1, line 155: This link shows the equation with D65 white point, however we are using D50. The original parameter is right.

lib/src/gainmapmath.cpp
```diff
@@ -66,28 +66,37 @@ void ShepardsIDW::fillShepardsIDW(float* weights, int incR, int incB) {
...
+// See https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#MODEL_YUV
+
 ////////////////////////////////////////////////////////////////////////////////
 // sRGB transformations
 
-// See IEC 61966-2-1/Amd 1:2003, Equation F.7.
-static const float kSrgbR = 0.2126f, kSrgbG = 0.7152f, kSrgbB = 0.0722f;
+// BT709 Luminance
+// See https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#PRIMARIES_BT709
+// BT709 chromaticity coordinates are also shared by sRGB
```
DichenZhang1, line 81: I would recommend keeping the original source. It's OK to keep both sources, so that this is a mark that we confirmed the numbers with two sources.

DichenZhang1: And same for the rest

lib/src/gainmapmath.cpp
DichenZhang1: The numbers are correct. Keep the original comment, but change the equation 
```
// See SMPTE EG 432-1, Equation G-7.
```

lib/src/gainmapmath.cpp
ram-mohan: Yes Dichen, made suggested changes.

lib/include/ultrahdr/icc.h
```diff
@@ -238,7 +244,7 @@ struct ICCHeader {
 class IccHelper {
  private:
   static constexpr uint32_t kTrcTableSize = 65;
-  static constexpr uint32_t kGridSize = 17;
+  static constexpr uint32_t kGridSize = 11;
```
DichenZhang1, line 247: I wonder where is this change coming from. The original value is aligned with skia: https://cs.android.com/android/platform/superproject/+/android14-qpr3-release:frameworks/native/libs/ultrahdr/include/ultrahdr/icc.h;l=218?q=kGridSize

lib/src/jpegr.cpp
```diff
@@ -262,11 +262,13 @@ uhdr_error_info_t JpegR::encodeJPEGR(uhdr_raw_image_t* hdr_intent, uhdr_raw_imag
...
-#else
-  UHDR_ERR_CHECK(convertYuv(sdr_intent_yuv, sdr_intent_yuv->cg, UHDR_CG_DISPLAY_P3));
-#endif
+//#if (defined(UHDR_ENABLE_INTRINSICS) && (defined(__ARM_NEON__) || defined(__ARM_NEON)))
+//  UHDR_ERR_CHECK(
+//      convertYuv_neon(sdr_intent_yuv, sdr_intent_yuv->cg, (uhdr_color_gamut_t)UHDR_CG_BT_601));
+//#else
+//  UHDR_ERR_CHECK(
+//      convertYuv(sdr_intent_yuv, sdr_intent_yuv->cg, (uhdr_color_gamut_t)UHDR_CG_BT_601));
+//#endif
```
DichenZhang1, line 271: I believe we intended to convert YUV. Is there any reason this was removed?

lib/src/jpegr.cpp
```diff
@@ -415,11 +464,14 @@ uhdr_error_info_t JpegR::convertYuv(uhdr_raw_image_t* image, uhdr_color_gamut_t
...
-      switch (dst_encoding) {
+      switch ((int)dst_encoding) {
+        case UHDR_CG_BT_601:
+          coeffs_ptr = &kYuvBt709ToBt601;
+          break;
         case UHDR_CG_BT_709:
           return status;
         case UHDR_CG_DISPLAY_P3:
-          coeffs_ptr = &kYuvBt709ToBt601;
+          coeffs_ptr = &kYuvBt709ToDisplayP3;
```
DichenZhang1, line 474: We intended to use BT601 YUV encoding here

DichenZhang1 (commented):
I suggest separating this change into two: changes before 10/29 to be the first one, and changes after 10/29 to be the second one. Thank you!

lib/include/ultrahdr/icc.h
ram-mohan: I have updated this basing on this,
https://cs.android.com/android/platform/superproject/main/+/main:external/skia/src/encode/SkICC.cpp;drc=1cce66c0004230c737a7ef3bbc1559015d83eaa6;l=719

currently the skICC.cpp and icc.cpp are different. I have taken this change, as it reduces the encoded size.

PR #228, opened 2024-08-06
API: change a parameter type from bool to int

ram-mohan:
Test: build


lib/src/ultrahdr_api.cpp
```diff
@@ -1540,7 +1540,7 @@ void uhdr_reset_decoder(uhdr_codec_private_t* dec) {
 }
 
 uhdr_error_info_t uhdr_enable_gpu_acceleration(uhdr_codec_private_t* codec,
```
DichenZhang1, line 1542: Curious why we want this change? Since it boolean type in uhdr_codec_private

ram-mohan: bool is not a defined data type in C 89/90 standard. Using this in C applications (image magick for instance) is causing build issues.

DichenZhang1 (approved):

PR #234, opened 2024-08-10
ossfuzz.sh: Opt out of shift sanitizer

harishdm:
Shift sanitizer flags a few failures that prevent further fuzzing with other undefined behavior sanitizers enabled.
Disable shift sanitizer till the library is updated to avoid these undefined shift sanitizer issues.


harishdm:
cc: @ram-mohan

DichenZhang1 (approved):

PR #235, opened 2024-08-10
Update ultrahdr_app.cpp

DichenZhang1:
Small fixes in the app


ram-mohan:
It seems most of the fixes are towards full range evaluations. Perhaps, we can squash all the 4 commits in to 1 commit and push again, git push --force origin  DichenZhang1-patch-3.

PR #238, opened 2024-08-15
add gamut space conversions in sample app.

ram-mohan:
If hdr intent and sdr intent are of different gamut space, during psnr computations it may be required to do gamut conversions. The logic necessary for the same is added.

Test: ./ultrahdr_app

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -60,6 +60,19 @@ const float BT2020RGBtoYUVMatrix[9] = {0.2627,
                                        (-0.6780 / 1.4746),
                                        (-0.0593 / 1.4746)};
 
+const float BT709ToBT2100RGB[9] = {0.627404, 0.329282, 0.043314, 0.069097, 0.919541,
+                                   0.011362, 0.016392, 0.088013, 0.895595};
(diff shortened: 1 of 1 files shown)


examples/ultrahdr_app.cpp
```diff
@@ -1302,6 +1317,67 @@ void UltraHdrAppInput::computeYUVSdrPSNR() {
...
+      float g0 =
+          float((rgbData[mDecodedUhdrRgbImage.stride[UHDR_PLANE_PACKED] * i + j] >> 10) & 0x3ff);
+      float b0 =
+          float((rgbData[mDecodedUhdrRgbImage.stride[UHDR_PLANE_PACKED] * i + j] >> 20) & 0x3ff);
+
+      r0 /= 1023.0f;
+      g0 /= 1023.0f;
+      b0 /= 1023.0f;
+
+      float r = coeffs[0] * r0 + coeffs[1] * g0 + coeffs[2] * b0;
```
DichenZhang1, line 1360: Thank you for the change!

I suppose this should happen in linear space. I.e. mDecodedUhdrRgbImage is HLG for example, need to first convert it to linear, then convert gamut, then convert back to HLG

ram-mohan: Sorry, i missed this part. This change wont work at all.

PR #239, opened 2024-08-19
Add gpu acceleration support for editor helper functions

vivekrj1806:
Test: ./ultrahdr_unit_test

Change-Id: Ifb4ada9156863bdbd3c5de54a07a1a4cdb1e2fb8

### DIFF
--- a/lib/include/ultrahdr/editorhelper.h
+++ b/lib/include/ultrahdr/editorhelper.h
@@ -119,16 +119,41 @@ extern void rotate_buffer_clockwise_neon(T* src_buffer, T* dst_buffer, int src_w
                                          int src_stride, int dst_stride, int degrees);
 #endif
 
+#ifdef UHDR_ENABLE_GLES
+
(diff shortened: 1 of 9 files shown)


lib/include/ultrahdr/editorhelper.h
```diff
@@ -19,6 +19,7 @@
 
 #include "ultrahdr_api.h"
 #include "ultrahdr/ultrahdrcommon.h"
+#include "ultrahdr/jpegr.h"
```
DichenZhang1, line 22: I think this is not needed. Could you double check?

vivekrj1806: This is needed since we are making use of isBufferDataContiguous function declared in jpegr.h

ram-mohan: @vivekrj1806  as editorhelper.h functionality is mostly isolated, it does not make sense for it to include jpegr.h, If you move is isBufferDataContiguous inside ultrahdrcommon.h then there is no need to include jpegr.h

vivekrj1806: Done.

PR #242, opened 2024-08-25
Improve rgb input handling

ram-mohan:
For rgb inputs, these are first converted to ycbcr inputs and stored internally. During gainmap generation, these are again converted back to rgb. These redundant color space conversions are removed.

fixed few compilation warnings

Test: ./ultrahdr_unit_test

### DIFF
--- a/lib/src/gpu/applygainmap_gl.cpp
+++ b/lib/src/gpu/applygainmap_gl.cpp
@@ -235,8 +235,9 @@ bool isBufferDataContiguous(uhdr_raw_image_t* img) {
     uint16_t* y = static_cast<uint16_t*>(img->planes[UHDR_PLANE_Y]);
     uint16_t* u = static_cast<uint16_t*>(img->planes[UHDR_PLANE_UV]);
     std::ptrdiff_t sz = u - y;
+    long pixels = img->w * img->h;
     return img->stride[UHDR_PLANE_Y] == img->w && img->stride[UHDR_PLANE_UV] == img->w &&
-           sz == img->w * img->h;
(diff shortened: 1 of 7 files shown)


lib/src/gpu/applygainmap_gl.cpp
```diff
@@ -235,8 +235,9 @@ bool isBufferDataContiguous(uhdr_raw_image_t* img) {
     uint16_t* y = static_cast<uint16_t*>(img->planes[UHDR_PLANE_Y]);
     uint16_t* u = static_cast<uint16_t*>(img->planes[UHDR_PLANE_UV]);
     std::ptrdiff_t sz = u - y;
+    long pixels = img->w * img->h;
```
DichenZhang1, line 238: Do we need to consider stride here?

DichenZhang1 (approved):

DichenZhang1:
Thank you for making this happen! A side question I assume we'll have a follow-up CL in the example app?

ram-mohan:
> Thank you for making this happen! A side question I assume we'll have a follow-up CL in the example app?

Sample app has changes that allow for rgb inputs. so no additional change is required.

lib/src/gpu/applygainmap_gl.cpp
ram-mohan: Here we are checking if the buffer data is contiguous. That is it has no horizontal stride (width and stride are same) and there is no gap between y/u/v buffers. To determine there is no gap between y,u,v buffers we are comparing u ptr - y ptr with total number of pixels. So width is appropriate here.

PR #253, opened 2024-08-30
configure hdr display max brightness

ram-mohan:
If the input color transfer is linear, there is no proper peak brightness configuration. Also hlg is not absolute standard allowing room for different peak brightness levels. This change addresses this.

Test: ./ultrahdr_unit_test


ram-mohan:
I think the changes made on decoder side need closer inspection.

PR #256, opened 2024-08-31
Modify few default encoding configurations

ram-mohan:
By default gainmap scale factor is set to 1, multi-channel gainmap encoding is enabled.

Test: ./ultrahdr_unit_test

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -256,9 +256,9 @@ class UltraHdrAppInput {
                    uhdr_color_transfer_t hdrTf = UHDR_CT_HLG, int quality = 95,
                    uhdr_color_transfer_t oTf = UHDR_CT_HLG,
                    uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102, bool isHdrCrFull = false,
-                   int gainmapScaleFactor = 4, int gainmapQuality = 85,
-                   bool enableMultiChannelGainMap = false, float gamma = 1.0f,
-                   bool enableGLES = false, uhdr_enc_preset_t encPreset = UHDR_USAGE_REALTIME,
+                   int gainmapScaleFactor = 1, int gainmapQuality = 95,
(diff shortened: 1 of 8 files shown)


examples/ultrahdr_app.cpp
```diff
@@ -256,8 +256,8 @@ class UltraHdrAppInput {
                    uhdr_color_transfer_t hdrTf = UHDR_CT_HLG, int quality = 95,
                    uhdr_color_transfer_t oTf = UHDR_CT_HLG,
                    uhdr_img_fmt_t oFmt = UHDR_IMG_FMT_32bppRGBA1010102, bool isHdrCrFull = false,
-                   int gainmapScaleFactor = 4, int gainmapQuality = 85,
-                   bool enableMultiChannelGainMap = false, float gamma = 1.0f,
+                   int gainmapScaleFactor = 1, int gainmapQuality = 85,
```
gregbenz, line 259: Thank you, I think this is great to help ensure high quality as the default (and users can of course use the flags to request lower quality where they feel it meets their size / quality objectives).

DichenZhang1 (approved):

lib/include/ultrahdr/jpegr.h
```diff
@@ -29,19 +29,19 @@
 namespace ultrahdr {
 
 // Default configurations
-// Map is quarter res / sixteenth size
-static const size_t kMapDimensionScaleFactorDefault = 4;
+// gainmap image downscale factor
+static const size_t kMapDimensionScaleFactorDefault = 1;
```
DichenZhang1, line 33: Actually can we keep this and `kUseMultiChannelGainMapDefault` to the old values? This library is also used by Android framework and Android is not ready for moving to this this high bar yet. On the other side since we have updated the values in the app so the app will have the new default values.

DichenZhang1: I mean can we only keep the original configuration `kMapDimensionScaleFactorDefault = 4`? The app's default is 1 and is passing it to here and will bypass 4

PR #261, opened 2024-09-03
Add java sample application for libuhdr

vivekrj1806:
Test: java -Djava.library.path=<path> -jar uhdr-java.jar

Change-Id: Icfd82938269cfceff2527847aa14cc0cc7492b3d


java/UltraHdrApp.java
```diff
@@ -0,0 +1,705 @@
...
+
+import com.google.media.codecs.ultrahdr.UltraHDRDecoder;
+import com.google.media.codecs.ultrahdr.UltraHDREncoder;
+import com.google.media.codecs.ultrahdr.UltraHDRDecoder.GainMapMetadata;
+import com.google.media.codecs.ultrahdr.UltraHDRDecoder.RawImage;
+
+/**
+ * Ultra HDR Encoding/Decoding Demo Application
+ */
+public class UltraHdrApp {
```
DichenZhang1, line 39: I assume this file can be directly imported by Java users, so maybe "UltraHdrApi" is better name as it aligns with what we currently have for C code ultrahdr_api.h. Other than this this change looks really nice! Thank you for making this happen!

vivekrj1806: This is a sample application and this is importing uhdr encoder and uhdr decoder APIs via the import statements.
import com.google.media.codecs.ultrahdr.UltraHDRDecoder
import com.google.media.codecs.ultrahdr.UltraHDREncoder
They both combine together to form ultrahdr_api.h
Do you want us to rename those classes?

DichenZhang1: OK do we have a "ultrahdr_api.h" - like java version file so that app developer can directly import that one, or it will be in the follow-up CL, or does app developer need to import UltraHDRDecoder and UltraHDREncoder separately?

vivekrj1806: App developer will need to import UltraHDRDecoder and UltraHDREncoder.

DichenZhang1: OK I see Thank you for clarification!

DichenZhang1 (approved):

PR #266, opened 2024-09-06
add API to set mastering display, target display peak brightness

ram-mohan:
add API to set mastering display, target display peak brightness

### DIFF
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -147,7 +147,8 @@ int GetCPUCoreCount() {
 
 JpegR::JpegR(void* uhdrGLESCtxt, size_t mapDimensionScaleFactor, int mapCompressQuality,
              bool useMultiChannelGainMap, float gamma, uhdr_enc_preset_t preset,
-             float minContentBoost, float maxContentBoost) {
+             float minContentBoost, float maxContentBoost, float masteringDispPeakBrightness,
+             float targetDispPeakBrightness) {
   mUhdrGLESCtxt = uhdrGLESCtxt;
   mMapDimensionScaleFactor = mapDimensionScaleFactor;
   mMapCompressQuality = mapCompressQuality;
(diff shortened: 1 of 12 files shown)


ram-mohan:
This is not a complete change as java interfaces also requires to be updated if we are introducing this API. This is added to get some comments from reviewers.

ram-mohan:
> This is not a complete change as java interfaces also requires to be updated if we are introducing this API. This is added to get some comments from reviewers.

java interfaces for this api are also added.

harishdm (approved):

lib/src/jpegr.cpp
```diff
@@ -498,6 +499,22 @@ uhdr_error_info_t JpegR::convertYuv(uhdr_raw_image_t* image, uhdr_color_gamut_t
   return status;
 }
 
+float JpegR::getMasteringDisplayMaxLuminance(uhdr_color_transfer_t transfer) {
+  switch (transfer) {
+    case UHDR_CT_LINEAR:
+      return mMaxDispBrightness;
+    case UHDR_CT_HLG:
+      return mMaxDispBrightness != -1.0f ? mMaxDispBrightness : kHlgMaxNits;
+    case UHDR_CT_PQ:
```
DichenZhang1, line 508: Does it make sense to make QP max brightness also configurable?

lib/src/jpegr.cpp
```diff
@@ -1479,16 +1501,17 @@ uhdr_error_info_t JpegR::applyGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_ima
 #endif
           }
 
-          rgb_hdr = rgb_hdr / display_boost;
           size_t pixel_idx = x + y * dest->stride[UHDR_PLANE_PACKED];
 
           switch (output_ct) {
             case UHDR_CT_LINEAR: {
+              rgb_hdr = rgb_hdr / gainmap_metadata->hdr_capacity_max;
```
DichenZhang1, line 1508: I feel with the logic the input `max_display_boost` will become unused in this pass. Also the logic in the OpenGL pass isn't updated

ram-mohan: Thank you. I have addressed this.

lib/src/jpegr.cpp
ram-mohan: Thank you. I have addressed this.

gregbenz:
Thank you for adding support, this is very helpful. Results look great.

DichenZhang1 (approved):

PR #273, opened 2024-09-16
add riscv64 compile support

Xeonacid:
Use rv64gc as a common baseline of Linux distros.


ram-mohan:
This looks to me like an incomplete change. It would be better to have a toolchain file for both 32 bit and 64 bit builds. Something like [this](https://cs.android.com/android/platform/superproject/main/+/main:external/libaom/build/cmake/toolchains/riscv-linux-gcc.cmake). Also update the building.md on how to compile for riscv targets.

Xeonacid:
> It would be better to have a toolchain file for both 32 bit and 64 bit builds.

I don't have a riscv32 environment to test and riscv32 is not well prepared for a Linux distro AFAIK, is it appropriate for me to add riscv64 changes only?

ram-mohan:
> > It would be better to have a toolchain file for both 32 bit and 64 bit builds.
> 
> I don't have a riscv32 environment to test and riscv32 is not well prepared for a Linux distro AFAIK, is it appropriate for me to add riscv64 changes only?

I think it should be fine. For whatever architecture the support is being added, lets add a toolchain and update the documentation

Xeonacid:
> This looks to me like an incomplete change. It would be better to have a toolchain file for both 32 bit and 64 bit builds. Something like [this](https://cs.android.com/android/platform/superproject/main/+/main:external/libaom/build/cmake/toolchains/riscv-linux-gcc.cmake). Also update the building.md on how to compile for riscv targets.

Updated.

ram-mohan (approved):
Thank you. Changes look good to me.

DichenZhang1 (approved):

PR #274, opened 2024-09-16
Reuse applyGainMap GL textures if effects are to be applied

vivekrj1806:
After applyGainMap, the frame buffers are copied to CPU. If image editing effects are to be applied, these are again copied to GPU. This consumes time and can be avoided. The current change defers GPU to CPU transfer to the end after all GL processing is done.

Test: ./ultrahdr_app -m 1 -j input_uhdr.jpg -o 1 -O 5
Test: ./ultrahdr_unit_test

Change-Id: I9b3e401027ab9d44e2c6aeb664fb10aee2496b52

### DIFF
--- a/lib/include/ultrahdr/ultrahdrcommon.h
+++ b/lib/include/ultrahdr/ultrahdrcommon.h
@@ -227,9 +227,10 @@ typedef struct uhdr_opengl_ctxt {
   EGLConfig mEGLConfig;   /**< EGL frame buffer configuration */
 
   // GLES Context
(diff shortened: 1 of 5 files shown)


DichenZhang1 (approved):

DichenZhang1:
Need to reopen this thread. We'll fix the failure and re-merge

ram-mohan:
> Need to reopen this thread. We'll fix the failure and re-merge

It is not clear why the build failed. Felt like a server side issue. re issued a pull request for this change.

PR #278, opened 2024-09-18
Add apis to access primary and gainmap compressed data

ram-mohan:
Also,
- fixed few api names for consistency.
- moved lib version macro to public interface
- documented version semantics inconsistencies

### DIFF
--- a/.github/workflows/cmake_linux.yml
+++ b/.github/workflows/cmake_linux.yml
@@ -43,6 +43,14 @@ jobs:
             cxx: clang++
             cmake-opts: '-DUHDR_BUILD_TESTS=1 -DUHDR_ENABLE_LOGS=1 -DUHDR_BUILD_DEPS=1 -DUHDR_SANITIZE_OPTIONS=address'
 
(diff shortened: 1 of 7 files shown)


DichenZhang1 (approved):

cgohlke:
Since `*gain_map*` was renamed to `*gainmap*`, wouldn't it make sense to also rename `UHDR_GAIN_MAP_IMG` to `UHDR_GAINMAP_IMG`?

ram-mohan:
> Since `*gain_map*` was renamed to `*gainmap*`, wouldn't it make sense to also rename `UHDR_GAIN_MAP_IMG` to `UHDR_GAINMAP_IMG`?

We renamed the functions from `*gain_map*` to `*gainmap*` because encoder setter functions are using *gainmap*` and decoder setter functions are using *gain_map*`. For consistency we used the same token everywhere. gainmap or gain_map are both fine but use the same everywhere was the idea. So for all functions and variables we have modified to gainmap. enum we left it as is.

Issue #279, opened 2024-09-18
Support encoding linear half float images


cgohlke:
As mentioned in discussion #143 and issue #268, it would be nice to be able to encode linear half float (16-bit float) images.

Passing a raw image with `img.fmt=UHDR_IMG_FMT_64bppRGBAHalfFloat`, `img.ct=UHDR_CT_LINEAR`, and `intent=UHDR_HDR_IMG` to the `uhdr_enc_set_raw_image` function currently (version 1.1.1) fails with `'unsupported input pixel format for hdr intent 4, expects one of {UHDR_IMG_FMT_24bppYCbCrP010, UHDR_IMG_FMT_32bppRGBA1010102}'`.

gregbenz:
concur, this would be very helpful to have 16-bit linear half float RGB support

isenberg:
Yes, would be good to have. I needed some format which can easily written by open source tools and used in my #292 RGB int 16+16+16, but RGBA float 16+16+16 or RGBA float 16+16+16+16 would also be fine for me.

Example to write it: magick infile.tif -depth 16 -define quantum:format=floating-point RGB:outfile.raw
Or for 16+16+16+16 with alpha channel use RGBA:outfile.raw.

gregbenz:
Thank you!

PR #292, opened 2024-09-28
added support for reading raw RGB 16+16+16

isenberg:
Currently as RGB input format only RGBA 10+10+10+2 bits is supported which can't be written by most open source tools due to the difference in depth for the alpha channel compared to the RGB channels.

This change adds reading support for RGB 16+16+16 bits which is a more frequently used format, for example using ImageMagick:

magick infile.tif -depth 16 RBG:outfile.raw

To enable reading RGB 16+16+16: -a 13
to use existing RGBA 10+10+10+2 code: -a 5

Disadvantage of this quick hack: Internally it still uses the 10+10+10+2 format for storage, which means the lower 6 bits of each channel are ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,17 @@
 ## This is an experimental fork ##
 Most likely you are looking for the original: https://github.com/google/libultrahdr
 
-This replaces HDR raw reading of format RGBA 10+10+10+2 by format RGB 16+16+16 which can be written by most other tools.
+This adds raw input image reading of format RGB 16+16+16 which can be written by most other tools.
 
-The upper 6 bits of each 16 bit channel are ignored in this version.
+The lower 6 bits of each 16 bit channel are ignored as internally it converts to 10+10+10+2 RGBA before HDR-JPEG conversion.
+
+RGB16+16+16 format is indicated by new command line parameter -a 13.
 
 Example:
-```magick imgfile.tif -depth 16 RGB:imgfile.raw```
+```
+magick imgfile.tif -depth 16 RGB:imgfile.raw
+ultrahdr_app -m 0 -p imgfile.raw -w RAWWIDTH -h RAWHEIGHT -a 13 -z outfile.jpg
+```
 
 ## Introduction
 
--- a/README.md
+++ b/README.md
@@ -1,3 +1,13 @@
+## This is an experimental fork ##
+Most likely you are looking for the original: https://github.com/google/libultrahdr
+
+This replaces HDR raw reading of format RGBA 10+10+10+2 by format RGB 16+16+16 which can be written by most other tools.
+
+The upper 6 bits of each 16 bit channel are ignored in this version.
+
+Example:
+```magick imgfile.tif -depth 16 RGB:imgfile.raw```
+
 ## Introduction
 
 libultrahdr is an image compression library that uses gain map technology
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * modified version supporting reading of raw RGB 16+16+16 with option -a 13, internally converted to 10+10+10+2
  */
 
 #ifdef _WIN32
@@ -141,7 +140,6 @@ class Profiler {
 
 /* reads RGB 16+16+16 bits which can be written by most tools
  * and converts to internal RGBA 10+10+10+2 bits by ignoring the lower 6 bits of each channel
- * example to create input raw file: magick imgin.tif -depth 16 RGB:imgout.raw
  * RGBA 10+10+10+2 format in memory: 32-bit little-endian: Red 9:0, Green 19:10, Blue 29:20, Alpha 31:30
  * 0: R7    ...    R0
  * 1: G5 ... G0 R9 R8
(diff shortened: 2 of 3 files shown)


gregbenz:
Might be worth having the tool issue a notice that the 16-bit input will be degraded to 10-bit, as this would not be suitable precision for data encoded with a linear EOTF.

isenberg:
Greg, I wondered if higher intensity resolution is possible without too much change, but as currently the only other possible RGB input is RGBA 10+10+10+2 bits, it's at least not worse. I would also be fine with using YUV input which is supported, but couldn't find an open source tool offering that very bit-format accepted here.

gregbenz:
Would 16-bit half float meet your needs? That would be high quality and enable linear encoding too. https://github.com/google/libultrahdr/issues/279. Perhaps that one update might solve both concerns.

Or do you need a need for 16-bit integer just for file format support?

isenberg:
Yes, half-float rgb 16+16+16 or rgba 16+16+16+16 would work as well as that can also be written by ImageMagick or other tools. Example:
magick infile.tif -depth 16 -define quantum:format=floating-point RGBA:outfile.raw

I cleaned up my patch regarding comments now and added a note to the -h output about the 16 to 10 conversion.

cgohlke:
My understanding that the `32bppRGBA1010102` format is not only about storage (that is easy to achieve with bit operations), but the color transfer must be "Hybrid log gamma" or "Perceptual Quantizer", not linear. No?

gregbenz:
That format is just raw numbers, no headers. The EOTF is specified separately to libultra (such as via command line flags with the demo app), so you should be able to encode and use linear with any supported raw input. 

Linear is one of the supported input options. We just need a higher precision input like half float in order to avoid quantization error that would likely occur with a linear EOTF if using 10-bit integer precision.

cgohlke:
Encoding linear `32bppRGBA1010102` succeeds for me but fails the roundtrip to linear `64bppRGBAHalfFloat`. The output looks like a color transfer function has been applied. I should double-check my code.

From the decoder documentation:

https://github.com/google/libultrahdr/blob/57247bf31e9c969d74d9d4da3b99d2bd7f6c454f/ultrahdr_api.h#L626-L630

ram-mohan:
@isenberg some changes that were made are reverted in later commits. May be you can squash all the changes to a single commit and do a force push. This will reduce the amount of changes to go through. Further image magick already has a coder for [ultrahdr](https://github.com/ImageMagick/ImageMagick/commit/8a83a2bb40c2c0b7446c15d566d0a958742290f5), with that in context may be this change is not needed?

cgohlke:
> fails the roundtrip to linear 64bppRGBAHalfFloat

Apologies. That actually works. It's the decoding to sRGB `32bppRGBA8888` that produces unexpected results:

UHDR encoded from linear RGBA1010102: uhdr
Expected output similar to this: expected

Anyway, this probably doesn't belong in this issue.

isenberg:
>  Further image magick already has a coder for [ultrahdr](https://github.com/ImageMagick/ImageMagick/commit/8a83a2bb40c2c0b7446c15d566d0a958742290f5), with that in context may be this change is not needed?

@cgohlke : That replaces my rgb16 patch indeed! I wasn't aware of that support in ImageMagick and will use that now.

Canceling my pull request...

ram-mohan:
> UHDR encoded from linear RGBA1010102: uhdr Expected output similar to this: expected
> 
> Anyway, this probably doesn't belong in this issue.

@cgohlke @gregbenz I would like to know few more details regarding this. So the input is 32bpp rgb1010102. From the aforth mentioned discussions I am assuming this is constructed by normalizing the rgb64 half float to 0-1 and then mapping it to 1024. Color transfer characteristics set to linear. Can you also tell me the peak luminance. For hlg it is 1000 nits and for PQ it is 10000 nits.

ram-mohan:
> replaces

@isenberg Please let me how i can help if you are facing any issue. Thanks

isenberg:
> @isenberg Please let me how i can help if you are facing any issue. Thanks

@ram-mohan , currently building it with --with-uhdr and trying out...

isenberg:
> > replaces
> 
> @isenberg Please let me how i can help if you are facing any issue. Thanks

Works fine! Great to see how ImageMagick now can write UHDR JPG with gain map using the libultrahdr from here.

In case others searching for the topic like try, build instructions: https://github.com/ImageMagick/ImageMagick/commit/4b85373fcd4de75118c0bb62266b66f8245906ec

Important on macOS with brew is to prepend the cmake with the following to use only libjpeg-turbo which replaces libjpeg: CFLAGS="-Iopt/homebrew/opt/libjpeg-turbo/include" LDFLAGS="-L/opt/homebrew/opt/libjpeg-turbo/lib"

In case

ram-mohan:
> > > replaces
> > 
> > 
> > @isenberg Please let me how i can help if you are facing any issue. Thanks
> 
> Works fine! Great to see how ImageMagick now can write UHDR JPG with gain map using the libultrahdr from here.
> 
> In case others searching for the topic like try, build instructions: [ImageMagick/ImageMagick@4b85373](https://github.com/ImageMagick/ImageMagick/commit/4b85373fcd4de75118c0bb62266b66f8245906ec)
> 
> Important on macOS with brew is to prepend the cmake with the following to use only libjpeg-turbo which replaces libjpeg: CFLAGS="-Iopt/homebrew/opt/libjpeg-turbo/include" LDFLAGS="-L/opt/homebrew/opt/libjpeg-turbo/lib"
> 
> In case

@isenberg All those things are no longer required.

1. install libjpeg (some thing like brew install jpeg-turbo / sudo apt install -y libjpeg-dev)
2. install libultrahdr
```
   git clone https://github.com/google/libultrahdr.git && cd libultrahdr && mkdir build && cd build
   cmake -G "Unix Makefiles" ../
   make
   sudo make install
```
3. clone image magick and configure with uhdr enabled
```
   git clone https://github.com/ImageMagick/ImageMagick.git && cd ImageMagick && mkdir build && cd build
   ../configure --with-uhdr && make
```

isenberg:
For further details about ImageMagick and UltraHDR I see #7198 is best.

PR #300, opened 2024-10-10
Make tonemapping and gainmap generation public

epeevs:
Change-Id: I34c515da63d97b334abdcaebb0eea23fed06667f

### DIFF
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -506,7 +506,8 @@ uhdr_error_info_t JpegR::compressGainMap(uhdr_raw_image_t* gainmap_img,
 uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_image_t* hdr_intent,
                                          uhdr_gainmap_metadata_ext_t* gainmap_metadata,
                                          std::unique_ptr<uhdr_raw_image_ext_t>& gainmap_img,
-                                         bool sdr_is_601, bool use_luminance) {
+                                         bool sdr_is_601, bool use_luminance,
+                                         bool use_multi_plane) {
   uhdr_error_info_t status = g_no_error;
 
   if (sdr_intent->fmt != UHDR_IMG_FMT_24bppYCbCr444 &&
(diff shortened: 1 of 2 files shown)


lib/src/jpegr.cpp
```diff
@@ -644,7 +645,8 @@ uhdr_error_info_t JpegR::generateGainMap(uhdr_raw_image_t* sdr_intent, uhdr_raw_
   }
 
   gainmap_img = std::make_unique<uhdr_raw_image_ext_t>(
-      mUseMultiChannelGainMap ? UHDR_IMG_FMT_24bppRGB888 : UHDR_IMG_FMT_8bppYCbCr400,
+      mUseMultiChannelGainMap ? UHDR_IMG_FMT_24bppRGB888 : (use_multi_plane) ?
+      UHDR_IMG_FMT_12bppYCbCr420 : UHDR_IMG_FMT_8bppYCbCr400,
```
ram-mohan, line 649: It is unclear why we are allocating more memory than required. generateGainMap() produces either rgb image or grayscale image. Here we are allocating for 420 format. Further in jpegencoderhelper.cpp, 420 inputs are marked as
```
    isGainMapImg = false; // <-- 420 inputs
    ...
    // this change skips this.
    if (isGainMapImg) {
      char comment[255];
      snprintf(comment, sizeof comment,
               "Source: google libuhdr v%s, Coder: libjpeg v%d, Attrib: GainMap Image",
               UHDR_LIB_VERSION_STR, JPEG_LIB_VERSION);
      jpeg_write_marker(&cinfo, JPEG_COM, reinterpret_cast<JOCTET*>(comment), strlen(comment));
    }
```

epeevs: Generally the proposed changes aim to extend support for UltraHDR gainmaps beyond the Jpeg case. In those scenarios where the gainmap is stitched in other containers (like ISOBMFF)  and encoded by other types of encoders (HEVC for example) neither grayscale nor rgb maybe available as input format. YUV420 inputs are far more popular and in some cases like the Android CDD the only ones mandated. So from this point of view allocating the final input buffer at this point skips unnecessary and potentially inefficient steps down the road.

epeevs: Never mind, I believe we can go with a more straightforward approach and avoid the yuv420 gainmap case here.

PR #302, opened 2024-10-11
Make tonemapping and gainmap generation public

epeevs:
Change-Id: I34c515da63d97b334abdcaebb0eea23fed06667f


DichenZhang1 (approved):

DichenZhang1:
A side note this structure doesn't look perfect but I'm OK with it for now. We do have plan to extend this library to HEIC/AVIF and by then it's better to move the gain map related methods out of jpegr class and make them sharable between JPEG and HEIF. (Backlogged on our roadmap)

PR #308, opened 2024-10-20
apply ootf and inv ootf for hlg transfers

ram-mohan:
This change also adds support for encoding linear half float hdr intent inputs


lib/src/gainmapmath.cpp
```diff
@@ -267,6 +286,24 @@ Color hlgInvOetfLUT(Color e_gamma) {
   return {{{hlgInvOetfLUT(e_gamma.r), hlgInvOetfLUT(e_gamma.g), hlgInvOetfLUT(e_gamma.b)}}};
 }
 
+Color hlgOotf(Color e, LuminanceFn luminance) {
+  float y = luminance(e);
+  return e * std::pow(y, 1.2f - 1.0f);
```
DichenZhang1, line 291: (This and following methods) I believe we should make a parameter `ootfLamda` instead of a magic number.

ram-mohan: Gamma = 1.2+0.42×log(LW/1000). As mastering display peak brightness by default is 1000 and this is not modifiable we are currently using 1.2 directly. 
.https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/libstagefright/renderfright/gl/ProgramCache.cpp;l=303

DichenZhang1 (commented):
Overall this change looks great, thank you for making this happen! I left one comment, and would suggest we separate the OOTF part with the rest, and merge them as two PRs. Thank you!

lib/src/gainmapmath.cpp
DichenZhang1: Yes but still recommend replacing this with a parameter, like what we did for `globalToneMap()` before instead of magic number flying around

ram-mohan: Yes, it is better that way, instead of using a magic constant. Done.

ram-mohan:
> Overall this change looks great, thank you for making this happen! I left one comment, and would suggest we separate the OOTF part with the rest, and merge them as two PRs. Thank you!

Dichen, as suggested i have kept only ootf changes as part of this pull request. Once this is merged. I will issue another request that address linear color transfer inputs.

DichenZhang1 (approved):

PR #316, opened 2024-11-03
fix possible overflows

ram-mohan:


DichenZhang1:
Hi Ram, I find this change has both fixing overflows and changing color coefficients, and both parts are not related I would recommend we separate those into two CLs

ram-mohan:
> Hi Ram, I find this change has both fixing overflows and changing color coefficients, and both parts are not related I would recommend we separate those into two CLs

Sure Dichen, i have update the change accordingly.

DichenZhang1 (approved):

PR #323, opened 2024-11-05
Add .gitignore

lhpqaq:


.gitignore
```diff
@@ -0,0 +1,28 @@
+**/**.ninja
```
ram-mohan, line 1: Is this not sufficient?

```
# build output directories
build
build*

# downloaded dependencies
third_party/googletest
third_party/turbojpeg
third_party/benchmark
tests/data
```

ram-mohan (commented):
These changes can be squashed with previous commit and can be force pushed.

.gitignore
lhpqaq: Because it is uncertain whether the build directory is named 'build' 😂

ram-mohan (approved):

.gitignore
ram-mohan: I have suggested these changes basing on this, https://github.com/google/jpegli/blob/main/.gitignore and https://github.com/google/heir/blob/main/.gitignore, thank you making them

DichenZhang1 (approved):

PR #325, opened 2024-11-09
add rgba8888 to yuv444 conversion using arm neon

ram-mohan:
This change implements a neon equivalent of
convert_raw_input_to_ycbcr(). However not all combinations are handled. Only rgba888 to yuv444 conversion is supported as it happens to be the most common use case.


harishdm (approved):

DichenZhang1 (approved):

lhpqaq:
@ram-mohan Hi, Could you provide some input materials and commands for testing?

PR #329, opened 2024-11-19
enable support for xmp, iso to be build time configurable

ram-mohan:
as per https://developer.android.com/media/platform/hdr-image-format implementations are expected to write both xmp and iso metadata by default.

Test: ./ultrahdr_app

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -40,7 +40,8 @@ cc_library {
         "lib/include",
     ],
     local_include_dirs: ["lib/include"],
-    cflags: ["-DUHDR_ENABLE_INTRINSICS"],
+    cflags: ["-DUHDR_ENABLE_INTRINSICS",
+        "-DUHDR_WRITE_XMP",],
     srcs: [
         "lib/src/icc.cpp",
         "lib/src/jpegr.cpp",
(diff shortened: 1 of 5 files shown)


koto:
Not that it should block the PR, but the images with both ISO and XMP gainmap metadata seem to be hitting Chromium bug [377620718](https://issues.chromium.org/issues/377620718).

gregbenz:
@koto While production Chrome won't render either (even Android-only, no ISO), Canary renders the ISO + XMP images just fine in my testing. As far as I can tell (in my limited testing), there is no value / need to delay this, as the issue with Chrome appears unrelated to ISO as far as I have seen. That issue seems to affect either encoding or both together, and ultimately means we need to hope the Chrome fix moves to production quickly. I have not tested Dev / Beta Chrome channels to see if they have the fix yet.

gregbenz:
@koto Latest production Chrome 131 has resolved the gain map bug (works on 131.0.6778.109 if not slightly older, resolved in the past week). I'm seeing Brave works as well (see test #8: https://gregbenzphotography.com/hdr/#tests)

ram-mohan:
@DichenZhang1 dichen, the changes are ready for review. thank you.

PR #330, opened 2024-11-20
Add support for RISC-V Vector extension

lhpqaq:
I am working to gradually add support for RISC-V Vector extension. Does the main branch have this plan?
I can follow your suggestion, such as changing to integer calculation like neon


lhpqaq:
> > I am working to gradually add support for RISC-V Vector extension. Does the main branch have this plan? I can follow your suggestion, such as changing to integer calculation like neon
> 
> Thank you for your interest. We are not working on this path currently. some practices I follow, generally during our development we try to make changes such that they work on both 32 and 64 bit variants. As integer operations tend to be faster than float arithmetic, while developing intrinsics, we chose fixed point implementation. We ensure the changes are clang-formatted so that the code style is uniform across files and sufficient testing needs to be done to ensure nothing trivial breaks.

Hi, I have implemented another version of RVV in #334, and I will continue to work on this PR in the future. Looking forward to your guidance

ram-mohan:
if this is no longer required, can this be closed?

PR #331, opened 2024-11-22
fix primaries used for display p3 gamut

ram-mohan:
- luminance function for p3 gamut is using weights of dci p3 instead of display p3. this is corrected. accordingly the gamut conversions involving display p3 are updated.
- increased precision of some constants used in csc

Test: ./ultrahdr_unit_test

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -89,7 +89,7 @@ void ShepardsIDW::fillShepardsIDW(float* weights, int incR, int incB) {
 // sRGB transformations
 
 // See IEC 61966-2-1/Amd 1:2003, Equation F.7.
-static const float kSrgbR = 0.2126f, kSrgbG = 0.7152f, kSrgbB = 0.0722f;
+static const float kSrgbR = 0.212639f, kSrgbG = 0.715169f, kSrgbB = 0.072192f;
 
 float srgbLuminance(Color e) { return kSrgbR * e.r + kSrgbG * e.g + kSrgbB * e.b; }
 
@@ -97,7 +97,7 @@ float srgbLuminance(Color e) { return kSrgbR * e.r + kSrgbG * e.g + kSrgbB * e.b
 // Uses the same coefficients for deriving luma signal as
 // IEC 61966-2-1/Amd 1:2003 states for luminance, so we reuse the luminance
 // function above.
-static const float kSrgbCb = 1.8556f, kSrgbCr = 1.5748f;
+static const float kSrgbCb = (2 * (1 - kSrgbB)), kSrgbCr = (2 * (1 - kSrgbR));
 
 Color srgbRgbToYuv(Color e_gamma) {
   float y_gamma = srgbLuminance(e_gamma);
@@ -121,7 +121,7 @@ float srgbInvOetf(float e_gamma) {
   if (e_gamma <= 0.04045f) {
     return e_gamma / 12.92f;
   } else {
-    return pow((e_gamma + 0.055f) / 1.055f, 2.4);
+    return pow((e_gamma + 0.055f) / 1.055f, 2.4f);
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/src/gainmapmath.cpp
```diff
@@ -85,28 +85,89 @@ void ShepardsIDW::fillShepardsIDW(float* weights, int incR, int incB) {
...
 
 // See IEC 61966-2-1/Amd 1:2003, Equation F.7.
-static const float kSrgbR = 0.2126f, kSrgbG = 0.7152f, kSrgbB = 0.0722f;
+static const float kSrgbR = 0.212639f, kSrgbG = 0.715169f, kSrgbB = 0.072192f;
 
 float srgbLuminance(Color e) { return kSrgbR * e.r + kSrgbG * e.g + kSrgbB * e.b; }
 
 // See ITU-R BT.709-6, Section 3.
-// Uses the same coefficients for deriving luma signal as
-// IEC 61966-2-1/Amd 1:2003 states for luminance, so we reuse the luminance
```
DichenZhang1, line 98: I think we can keep the notes

lib/src/gainmapmath.cpp
```diff
@@ -85,28 +85,89 @@ void ShepardsIDW::fillShepardsIDW(float* weights, int incR, int incB) {
...
-// function above.
-static const float kSrgbCb = 1.8556f, kSrgbCr = 1.5748f;
+static const float kSrgbCb = (2 * (1 - kSrgbB)), kSrgbCr = (2 * (1 - kSrgbR));
 
 Color srgbRgbToYuv(Color e_gamma) {
   float y_gamma = srgbLuminance(e_gamma);
   return {{{y_gamma, (e_gamma.b - y_gamma) / kSrgbCb, (e_gamma.r - y_gamma) / kSrgbCr}}};
 }
 
-// See ITU-R BT.709-6, Section 3.
```
DichenZhang1, line 107: Same

lib/src/gainmapmath.cpp
```diff
@@ -190,45 +251,18 @@ Color p3YuvToRgb(Color e_gamma) {
...
-// Simplify:
-//   G = Y * ((1 - kBt2100R - kBt2100B) / kBt2100G)
-//     + U * (kBt2100B * kBt2100Cb / kBt2100G)
-//     + V * (kBt2100R * kBt2100Cr / kBt2100G)
-//
-// We then get the following coeficients for calculating G from YUV:
-//
-// Coef for Y = (1 - kBt2100R - kBt2100B) / kBt2100G = 1
-// Coef for U = kBt2100B * kBt2100Cb / kBt2100G = kBt2100GCb = ~0.1645
-// Coef for V = kBt2100R * kBt2100Cr / kBt2100G = kBt2100GCr = ~0.5713
```
DichenZhang1, line 230: same

DichenZhang1 (commented):
LGTM. Thank you!

lib/src/gainmapmath.cpp
```diff
@@ -600,41 +603,43 @@ void putYuv444Pixel(uhdr_raw_image_t* image, size_t x, size_t y, Color& pixel) {
...
+  return {{{1.660491f * e.r + -0.587641f * e.g + -0.07285f * e.b,
+            -0.124551f * e.r + 1.1329f * e.g + -0.008349f * e.b,
+            -0.018151f * e.r + -0.100579f * e.g + 1.11873f * e.b}}};
 }
 
 Color bt2100ToP3(Color e) {
-  return {{{1.34369f * e.r + -0.28223f * e.g + -0.06135f * e.b,
-            -0.06533f * e.r + 1.07580f * e.g + -0.01051f * e.b,
-            0.00283f * e.r + -0.01957f * e.g + 1.01679f * e.b}}};
+  return {{{1.343578f * e.r + -0.282179f * e.g + -0.061399f * e.b,
```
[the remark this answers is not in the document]
ram-mohan: the matrix multiplication seems to be wrong. The individual matrices are same as the ones i am using.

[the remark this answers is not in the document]
DichenZhang1: OK I re-did the math with gemini and this time the result matches with yours. My previous result with chatGPT was wrong. Sorry for the confusion.

DichenZhang1 (approved):

PR #332, opened 2024-11-23
use cicp tag info if present while parsing for gamut space

ram-mohan:
color space of the image can be signalled via cicp tag in icc. While parsing for gamut space of the image, use the information present in this tag as well (if available).


DichenZhang1 (approved):
Could you replace `total_length` with `kCicpTagSize` at line 276?

ram-mohan:
> Could you replace `total_length` with `kCicpTagSize` at line 276?

Done dichen

PR #333, opened 2024-11-24
improve gainmap computation logic for dark pixels

ram-mohan:
while computing gain for near blacks or pixels with channel values zero, sdr_offset is used. This value is extremely small and can result large gainmap coefficient. During encode-decode process, if this sdr pixel flips to zero, after applying gainmap the hdr pixel will blow out. It is best if the gainmap computation logic is desensitized for dark pixels.

also, conversion from wider to narrower gamut space can cause overrange values. clip only negative excursions so that domain of computeGain remains positive.

Test: ./ultrahdr_app

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -782,12 +782,14 @@ uint8_t encodeGain(float y_sdr, float y_hdr, uhdr_gainmap_metadata_ext_t* metada
 }
 
 float computeGain(float sdr, float hdr) {
-  if (sdr == 0.0f) return 0.0f;  // for sdr black return no gain
-  if (hdr == 0.0f) {  // for hdr black, return a gain large enough to attenuate the sdr pel
-    float offset = (1.0f / 64);
-    return log2(offset / (offset + sdr));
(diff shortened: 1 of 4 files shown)


lib/src/gainmapmath.cpp
```diff
@@ -782,12 +782,14 @@ uint8_t encodeGain(float y_sdr, float y_hdr, uhdr_gainmap_metadata_ext_t* metada
 }
 
 float computeGain(float sdr, float hdr) {
-  if (sdr == 0.0f) return 0.0f;  // for sdr black return no gain
-  if (hdr == 0.0f) {  // for hdr black, return a gain large enough to attenuate the sdr pel
-    float offset = (1.0f / 64);
-    return log2(offset / (offset + sdr));
+  float gain = log2((hdr + 1e-7f) / (sdr + 1e-7f));
```
DichenZhang1, line 785: I would recommend making `1e-7f` a global constant since it has been used in multiple places.

DichenZhang1 (approved):

lib/src/gainmapmath.cpp
ram-mohan: Done dichen.

PR #335, opened 2024-12-03
signal alternate image color space in gainmap image

ram-mohan:
If sdr intent and hdr intent gamut space are not identical, it is required to send hdr intent color profile as part of gainmap image. This is addressed.

Change-Id: I0155a1dc043f3c3e0493e04e5a6eabe160756476

### DIFF
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -49,7 +49,8 @@ namespace ultrahdr {
 uhdr_error_info_t applyGainMapGLES(uhdr_raw_image_t* sdr_intent, uhdr_raw_image_t* gainmap_img,
                                    uhdr_gainmap_metadata_ext_t* gainmap_metadata,
                                    uhdr_color_transfer_t output_ct, float display_boost,
-                                   uhdr_raw_image_t* dest, uhdr_opengl_ctxt_t* opengl_ctxt);
+                                   uhdr_color_gamut_t sdr_cg, uhdr_color_gamut_t hdr_cg,
+                                   uhdr_opengl_ctxt_t* opengl_ctxt);
 #endif
 
 // Gain map metadata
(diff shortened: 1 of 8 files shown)


lib/src/jpegr.cpp
```diff
@@ -494,7 +495,9 @@ uhdr_error_info_t JpegR::convertYuv(uhdr_raw_image_t* image, uhdr_color_gamut_t
 
 uhdr_error_info_t JpegR::compressGainMap(uhdr_raw_image_t* gainmap_img,
                                          JpegEncoderHelper* jpeg_enc_obj) {
-  return jpeg_enc_obj->compressImage(gainmap_img, mMapCompressQuality, nullptr, 0);
+  std::shared_ptr<DataStruct> icc = IccHelper::writeIccProfile(gainmap_img->ct, gainmap_img->cg);
```
DichenZhang1, line 498: I think we need to hide this part under iso metadata feature flag.

ram-mohan: Done

lib/src/jpegr.cpp
```diff
@@ -494,6 +495,11 @@ uhdr_error_info_t JpegR::convertYuv(uhdr_raw_image_t* image, uhdr_color_gamut_t
 
 uhdr_error_info_t JpegR::compressGainMap(uhdr_raw_image_t* gainmap_img,
                                          JpegEncoderHelper* jpeg_enc_obj) {
+  if (!kWriteXmpMetadata) {
+    std::shared_ptr<DataStruct> icc = IccHelper::writeIccProfile(gainmap_img->ct, gainmap_img->cg);
```
DichenZhang1, line 499: Thank you!

There is another flag in the iso metadata `useBaseColorSpace`. The gain map / icc logic should also be hidden when that flag is false, and the metadata encode / decode part also needs modification to support this feature.

ram-mohan: useBaseColorSpace is not configurable on the encoding side. as hdr intent is being converted to sdr intent and sdr image is base image, gain map application space shall always be base color space. hence useBaseColorSpace is always true on encoding side.
The change associated with selecting useBaseColorSpace dynamically are in the follow up CL. This change just signals hdr intent color space in gainmap image icc. In the next CL as useBaseColorSpace is selected dynamically and the encode / decode part is also modified. Further this dynamic selection is disabled if kWriteXmpMetadata is enabled.

DichenZhang1: OK

DichenZhang1 (approved):

lib/src/jpegr.cpp
ram-mohan: Dichen, this pr, #336, perhaps addresses some of the concerns you raised.

PR #337, opened 2024-12-06
enable support for handling gainmap multichannel metadata

ram-mohan:


gregbenz:
Great to see support expanding here. I added a few comments / questions on assumptions.

DichenZhang1 (approved):

PR #338, opened 2024-12-13
update weights used in rgb to yuv conversion equations

ram-mohan:
Also simplify sdr rgb jpeg encoding.


gregbenz:
Using BT.601 explicitly for JPEG and converting RGB directly to it avoids an extra conversion. Could that improvement be kept separate from changing how raw Display P3 YUV input is interpreted? The existing code uses BT.601 coefficients for P3 to match DataSpace. Changing those coefficients would appear to change the colors produced from existing inputs that follow that convention.

Keeping the current behavior, or letting callers explicitly choose the conversion matrix, would avoid that compatibility problem. Tests covering existing P3 YUV inputs and JPEG encoding from both RGB and YUV would help check the full pipeline, beyond the updated matrix tests. The standard and NEON-optimized paths should also produce matching results within the expected rounding tolerance.

PR #340, opened 2024-12-14
icc: update tone mapping

ram-mohan:
- update tone mapping to reflect the implementation in api-0.
- update the icc profile to output results in xyz D50 instead of Lab.
- update the icc profile to output tonemap results in api-0 only.
- add chad tag.

Test: ./ultrahdr_unit_test

### DIFF
--- a/lib/include/ultrahdr/icc.h
+++ b/lib/include/ultrahdr/icc.h
@@ -121,6 +121,7 @@ static constexpr uint32_t kTAG_wtpt = SetFourByteTag('w', 't', 'p', 't');
 static constexpr uint32_t kTAG_rTRC = SetFourByteTag('r', 'T', 'R', 'C');
 static constexpr uint32_t kTAG_gTRC = SetFourByteTag('g', 'T', 'R', 'C');
 static constexpr uint32_t kTAG_bTRC = SetFourByteTag('b', 'T', 'R', 'C');
+static constexpr uint32_t kTAG_chad = SetFourByteTag('c', 'h', 'a', 'd');
 static constexpr uint32_t kTAG_cicp = SetFourByteTag('c', 'i', 'c', 'p');
 static constexpr uint32_t kTAG_cprt = SetFourByteTag('c', 'p', 'r', 't');
 static constexpr uint32_t kTAG_A2B0 = SetFourByteTag('A', '2', 'B', '0');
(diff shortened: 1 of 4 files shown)


lib/include/ultrahdr/icc.h
```diff
@@ -153,6 +155,12 @@ static constexpr Matrix3x3 kRec2020 = {{
     {-0.00193139f, 0.0299794f, 0.797162f},
 }};
 
+static constexpr Matrix3x3 bradford = {{
```
DichenZhang1, line 158: 1. Better to provide data source as a comment.
2. Maybe append "f" to indicate the data type
3. I think our data selection of kDisplayP3 has already taken the white point shifting into account, this may not be needed.

lib/src/icc.cpp
```diff
@@ -432,49 +344,30 @@ std::shared_ptr<DataStruct> IccHelper::writeIccProfile(uhdr_color_transfer_t tf,
...
+    // Use empty data to indicate that the entry should use the previous tag's
+    // data.
+    tags.emplace_back(kTAG_gTRC, make_empty());
+    // Use empty data to indicate that the entry should use the previous tag's
+    // data.
+    tags.emplace_back(kTAG_bTRC, make_empty());
   }
 
+  // Chroma adaptation matrix
+  tags.emplace_back(kTAG_chad, write_chad_tag());
```
DichenZhang1, line 369: 1. Does it need to happen for all cases or just display P3
2. Even for display P3 I think our data selection of kDisplayP3 has already taken the white point shifting into account, so this may not be needed.

lib/include/ultrahdr/icc.h
ram-mohan: appended 'f' to float constants and indicated the source from which the constants are derived.

gregbenz:
Aligning the ICC fallback with API-0 would improve consistency, but two details may need correction: PQ still declares Lab PCS although the new transform produces XYZ D50, and `compute_lut_entry()` passes `false` to `globalTonemap()` after inverse PQ/HLG conversion, whereas API-0 treats those inputs as normalized.

Evaluating the tone-mapping stage at peak white gives approximately 0.5002 for PQ and 0.5206 for HLG instead of 1.0, before ICC serialization. Could these be corrected and covered by PQ/HLG comparisons against API-0, including validation through an independent ICC consumer?

PR #347, opened 2025-01-10
Revert "bump to 1.3.1"

DichenZhang1:
Reverts google/libultrahdr#339


ram-mohan:
@DichenZhang1 this is not required now

PR #359, opened 2025-05-27
Add FreeBSD to recognized platforms in main CMakeLists.txt

ricera:
This allows libultrahdr to build on FreeBSD.

It appears to work fine (I was able to generate a gain mapped jpeg image that looked correct when I viewed it in Chrome), so I don't think there's anything else that needs to change for this to work.

I don't know if it's worth adding an entry for it to docs/building.md since I doubt there will be official support for it.


DichenZhang1 (approved):

ricera:
Thanks!

PR #363, opened 2025-07-14
Add missing header for GCC 15

dbermond:
The C++ header [`<cstdint>`](https://en.cppreference.com/w/cpp/header/cstdint.html) is explicitly required to be included with GCC 15 and beyond when using [fixed width integer types](https://en.cppreference.com/w/cpp/types/integer.html).

For more details, see the guide about [porting to GCC 15](https://gcc.gnu.org/gcc-15/porting_to.html#cxx).

### DIFF
--- a/lib/include/ultrahdr/ultrahdrcommon.h
+++ b/lib/include/ultrahdr/ultrahdrcommon.h
@@ -24,6 +24,7 @@
 #include <GLES3/gl3.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


DichenZhang1 (approved):

cmahnke:
Since this is required to build on a recent Alpine (and certainly others which ship GCC 15), maybe a backport to a 1.4.1 would be possible?

PR #368, opened 2025-08-31
add support for probe mode

### LINKED ISSUE #362: Request: add option in CLI (demo app) to validate a source file as a supported gain map
It would be very helpful if a flag could be added to the CLI tool (demo app) which could be used to validate if a given file was a gain map with encoding supported by libultrahdr. This would be very helpful for a number of reasons, including to be able to quickly determine if processing of an image should be done via ultrahdr_app (if supported) or passed off to another tool (as there are many valid JPGs which are not gain maps).

something like: **ultrahdr_app -v myFile.jpg**

The result would just be a simple string indicating support or not. Basically, it would indicate if decoding would be possible, without consuming the time and resources to actually decode the image.

ram-mohan:
This change provides an option to print gainmap metadata without completely decoding the image.

fixes #362

Test: ./ultrahdrapp -m 1 -j out.jpeg -P

### DIFF
--- a/examples/ultrahdr_app.cpp
+++ b/examples/ultrahdr_app.cpp
@@ -91,6 +91,7 @@ int getopt_s(int argc, char* const argv[], char* ostr) {
   }
   if (oindex[1] != ':') {
     optarg_s = nullptr;
+    ++optind_s;
     return optopt_s;
   }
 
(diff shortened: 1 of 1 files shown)


ram-mohan:
@gregbenz probe() is the first stage of decoding process. In this stage, the application checks for presence of primary image, gainmapimage and gainmap metadata. If these are present then the application continues decoding. So i think this is a good starting point to check if the image is a valid uhdr image.

gregbenz:
Thank you, that's helpful. Possible to expose that with a switch in the CLI
app?

gregbenz:
Great! 

Is this supported to just probe (test if file is valid or not) without creating a new raw file?
./ultrahdrapp -j out.jpeg -P

I would generally assume the goal for using probe is just a test and not output (as requesting normal decode would be expected to throw an error if the target source image is not a valid gain map).

PR #376, opened 2025-12-09
Add ppc64le and s390x to recognized architectures

kleisauke:
See for example:
https://github.com/kleisauke/libvips-packaging/actions/runs/20065009063/job/57551203630
https://github.com/kleisauke/libvips-packaging/actions/runs/20065009063/job/57551203641

Note that I considered adding architecture-specific compile flags, similar to what's done here:
https://github.com/google/libultrahdr/blob/8cbc983d2f6c2171af5cbcdb8801102f83fe92ab/CMakeLists.txt#L295-L323

However, I don't think that makes much sense, as Linux distributions usually handle this themselves.


kleisauke:
Alternatively, here's a changeset that removes the platform and architecture detection logic (including the arch-specific flags):
https://github.com/google/libultrahdr/compare/main...kleisauke:remove-platform-arch-detect

For example, this allows building for a Raspberry Pi Zero W (ARMv6). Without this change, the build fails with:
```
cc1plus: warning: switch '-mcpu=arm1176jzf-s' conflicts with '-march=armv7-a' switch
Assembler messages:
Error: unknown floating point format `neon-vfpv3'
```
(and this is just one case, there are likely better examples of why injecting those flags is not advisable)

ram-mohan (approved):

kmilos:
> there are likely better examples of why injecting those flags is not advisable

FWIW, build on Fedora 43 w/ GCC 15 fails w/
```
/usr/bin/ld: /tmp/ccdT2utQ.ltrans0.ltrans.o: in function `std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release()':
/usr/include/c++/15/ext/atomicity.h:68:(.text+0x2ce8): undefined reference to `__atomic_fetch_add_4'
/usr/bin/ld: /usr/include/c++/15/ext/atomicity.h:68:(.text+0x2d3c): undefined reference to `__atomic_fetch_add_4'
```
presumably because the forced `-march=i386` overrides the distribution set `-march=i686` CFLAGS...? @topazus

Interestingly, linking luckily works out w/ the newer GCC 16 toolchain in Fedora 44 - I guess it now somehow figures out better it actually needs to link `libatomic` (although both have the `-Wl,--as-needed` flag)... (Note that in reality `libatomic` would be unnecessary in this case for i686.)

topazus:
> Interestingly, linking luckily works out w/ the newer GCC 16 toolchain in Fedora 44 - I guess it now somehow figures out better it actually needs to link `libatomic` (although both have the `-Wl,--as-needed` flag)... (Note that in reality `libatomic` would be unnecessary in this case for i686.)

It may be related to automatic linking of libatomic in GCC 16. ref: https://github.com/gcc-mirror/gcc/commit/e63cf4b130b86dd7dde1bf499d3d40faca10ea2e

kmilos:
@topazus I feel the alternative of [removing `-march=i386` for all Fedora versions might be better than explicitly asking for `-latomic`](https://stackoverflow.com/a/47498167) conditionally?

kleisauke:
Closing in favor of #383.

PR #381, opened 2026-01-25
Add UHDR_ENABLE_SYSTEM_GTEST option to use system GoogleTest

topazus:
Add UHDR_ENABLE_SYSTEM_GTEST option to choose to use system gtest dependency or not, default is FALSE. It will help do tests simpler in Linux packaging, which it prefers using system dependencies.


ram-mohan:
This was done long back but was reverted due to problems. Can you please take a look at this, https://github.com/google/libultrahdr/pull/83

PR #383, opened 2026-01-28
Remove platform and architecture detection logic

kleisauke:
Also drop arch-specific compile and optimization flags to ensure packaging friendliness.

Context: #376.


gregbenz:
Improving cross-platform builds is a great direction.

On Apple Silicon, setting `CMAKE_OSX_ARCHITECTURES=x86_64` can leave `CMAKE_SYSTEM_PROCESSOR=arm64`, so the new condition selects ARM sources for an Intel build. A small CMake probe using that condition fails on `arm_neon.h`.

Could the NEON source selection also account for `CMAKE_OSX_ARCHITECTURES`, while still letting the toolchain choose the architecture flags?

kleisauke:
@gregbenz Fixed with commit 7af35883ee575266cc43bbd05686734ebd5b1a82. Thanks for your review!

gregbenz:
@kleisauke Thanks, looks good!

DichenZhang1 (approved):

PR #385, opened 2026-03-13
Decode Apple gain maps

### LINKED ISSUE #195: Request: Decode support for images with Apple gain map format
It would be helpful to have support to decode images captured with an iPhone, which includes JPG and HEIC images with the Apple gain map format. This would help to transcode images captured with an iPhone (such as for uploading to web services).

If you attempt to decode such an image now, libultrahdr_app will throw "xml parse error, could not find attribute hdrgm:Version".

Support for decoding is all that is needed. Output in the Android format (and ultimately with the ISO gain map spec when finalized) is sufficient.

vrabaud:
Code Gemini-helped by, and examples taken from the Apache 2.0 libavif: https://github.com/AOMediaCodec/libavif/pull/2960

This fixes https://github.com/google/libultrahdr/issues/195


lib/include/ultrahdr/jpegr.h
DichenZhang1: Better to also add exif here

DichenZhang1 (approved):
In general this looks good to me. Could you please add a note on Apple's gain map description. Is this the right one: https://developer.apple.com/documentation/appkit/applying-apple-hdr-effect-to-your-photos

lib/include/ultrahdr/jpegr.h
vrabaud: Done

vrabaud:
I added the Apple gain map link. Though the format was reverse engineered by https://github.com/grapeot/AppleJPEGGainMap

PR #388, opened 2026-04-29
fuzzers: add new harness to be consumed by OSS-Fuzz

DavidKorczynski:
Adds three new fuzzing harnesses to be consumed by OSS-Fuzz. The goal is to increase the current code coverage
(https://storage.googleapis.com/oss-fuzz-coverage/libultrahdr/reports/20260427/linux/src/report.html) I have verified locally that the coverage improves significantly where it gets above 50% in the lib/ folder from a 30 second run verssu the 33% in the current oss-fuzz coverage report.


gregbenz:
These additional fuzz tests look useful. Could the ISO metadata test check that decoding succeeded before converting the result? A truncated input can leave `decodedMetadata` uninitialized, so the test could report an error caused by the test itself.

For the editor test, choosing the operation and dimensions before reading the pixel data would help exercise more cases. If the pixel data consumes all remaining input bytes, the test defaults to a 90° rotation. Adding P010 and filling each supported image plane with the correct number of bytes would also improve HDR coverage. An updated coverage report for each test would help show what these changes exercise.

PR #392, opened 2026-05-18
gainmapmath: fix OOB reads in encoder pixel processing on odd-dimension images

jortles:
## Summary

- **Bug 1 — `getP010Pixel` UV overflow on odd width**: The interleaved chroma index `(x & ~0x1) + 1` reads 1 element past the chroma plane when image width is odd and `x` is the last pixel in a row. Fix: clamp `chroma_x` so the V sample stays in bounds.
- **Bug 2 — `convert_raw_input_to_ycbcr` row/column overflow on odd height/width**: The 2×2 block loops access `row[i+1]` and `col[j+1]` unconditionally, overflowing when height or width is odd. Fix: change loop bounds from `i < h` to `i + 1 < h` (and `j + 1 < w`) so incomplete trailing blocks are skipped. Applied to both the RGBA1010102 and RGBA8888 paths.

The public encoder API (`uhdr_enc_set_raw_image`) already rejects odd dimensions for P010/YCbCr420, so these are defense-in-depth fixes for internal callers that bypass that validation.

Found by AFL++ fuzzing with a custom pixel-format harness.

## Test plan

- [ ] Verify existing unit tests pass (no behavior change for even-dimension images)
- [ ] Confirm ASan-instrumented build no longer triggers OOB on odd-dimension inputs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -429,7 +429,11 @@ Color getP010Pixel(uhdr_raw_image_t* image, size_t x, size_t y) {
   size_t chroma_stride = image->stride[UHDR_PLANE_UV];
 
   size_t pixel_y_idx = y * luma_stride + x;
-  size_t pixel_u_idx = (y >> 1) * chroma_stride + (x & ~0x1);
+  size_t chroma_x = x & ~(size_t)0x1;
+  // Clamp chroma_x so that the interleaved V sample (at chroma_x + 1) stays within the
+  // chroma row.  Without this, odd image widths cause a 1-element OOB read.
+  if (chroma_x + 1 >= image->w) chroma_x = image->w >= 2 ? image->w - 2 : 0;
+  size_t pixel_u_idx = (y >> 1) * chroma_stride + chroma_x;
   size_t pixel_v_idx = pixel_u_idx + 1;
 
   uint16_t y_uint = luma_data[pixel_y_idx] >> 6;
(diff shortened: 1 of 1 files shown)


lib/src/gainmapmath.cpp
```diff
@@ -1316,8 +1320,8 @@ std::unique_ptr<uhdr_raw_image_ext_t> convert_raw_input_to_ycbcr(uhdr_raw_image_
     uint16_t* uData = static_cast<uint16_t*>(dst->planes[UHDR_PLANE_UV]);
     uint16_t* vData = uData + 1;
 
-    for (size_t i = 0; i < dst->h; i += 2) {
-      for (size_t j = 0; j < dst->w; j += 2) {
+    for (size_t i = 0; i + 1 < dst->h; i += 2) {
```
DichenZhang1, line 1323: Thank you for pointing this out!

Generally this fixed the OOB reading on odd dimensions, however, when the boundary is odd, the last row/column is going to be ignored, leading to (0,0,0) value in YUV which maps to a green-ish color in the last row/column.

A better solution can be replicating the edge values

```
    for (size_t i = 0; i < dst->h; i += 2) {
      for (size_t j = 0; j < dst->w; j += 2) {
        size_t i0 = i;
        size_t i1 = std::min(i + 1, dst->h - 1);
        size_t j0 = j;
        size_t j1 = std::min(j + 1, dst->w - 1);

        // Read pixels using clamped coordinates (i0/i1, j0/j1)
        pixel[0] = read_pixel(rgbData, i0, j0);
        pixel[1] = read_pixel(rgbData, i0, j1); // Replicates pixel[0] if j is odd last col
        pixel[2] = read_pixel(rgbData, i1, j0); // Replicates pixel[0] if i is odd last row
        pixel[3] = read_pixel(rgbData, i1, j1);
        
        // ... perform conversion ...

        // Write Y (only for valid pixels)
        yData[stride * i + j] = pixel[0].y;
        if (j + 1 < dst->w) yData[stride * i + j + 1] = pixel[1].y;
        if (i + 1 < dst->h) {
          yData[stride * (i + 1) + j] = pixel[2].y;
          if (j + 1 < dst->w) yData[stride * (i + 1) + j + 1] = pixel[3].y;
        }

        // Write UV (always write, as we calculated it using replicated pixels)
        uData[chroma_stride * (i / 2) + j] = pixel[0].u;
        vData[chroma_stride * (i / 2) + j] = pixel[0].v;
      }
    }
```

DichenZhang1 (changes requested):

PR #394, opened 2026-05-27
Fix JPEG segment ordering per JFIF and CIPA DC-007 standards

minatoAI:
## Summary

The `appendGainMap` function builds JPEG output with incorrect APP marker ordering:

1. **JFIF (APP0)** is written after XMP, ISO, MPF and ICC segments — it should be the first APP marker after SOI (per JFIF/JPEG specification)
2. **MPF (APP2)** is written before JFIF and ICC, far from SOS — it should be placed right before SOS (per CIPA DC-007 Multi-Picture Format standard)

This causes HDR decoders that follow the standard ordering (e.g., Xiaohongshu/小红书) to fail to recognize the Ultra HDR gain map.

## Changes

1. **lib/src/jpegr.cpp**: Extract APP0(JFIF) from the base JPEG data and write it immediately after SOI. Parse the base JPEG marker segments (DQT, SOF, DHT) and write them before MPF. Write MPF right before SOS per CIPA DC-007.

2. **lib/src/jpegrutils.cpp**: Add `rdf:about=""` attribute and `hdrgm:OplusScale="-1"` field to the gain map XMP metadata block, matching the Android Ultra HDR specification (developer.android.com/media/platform/hdr-image-format).

3. **CMakeLists.txt**: Add `-DWITH_CRT_DLL=ON` to turbojpeg's ExternalProject CMAKE_ARGS when building on MSVC, to prevent CRT mismatch when the parent project uses dynamic CRT (/MD).

## Testing

Verified that the fix produces a JPEG segment order identical to camera-captured Ultra HDR images (e.g., OPPO Find X8 Ultra), and the reordered file is correctly recognized as HDR by the Xiaohongshu platform.

### DIFF
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -1295,6 +1295,7 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
   size_t sos_offset = 0;
   while (base_pos < base_size) {
       if (base_data[base_pos] != 0xFF) break;
+      if (base_pos + 1 >= base_size) break;  // guard against truncated data
       uint8_t marker = base_data[base_pos+1];
       if (marker == 0xDA) {            // SOS — stop here, MPF goes before this
           sos_offset = base_pos;
@@ -1303,7 +1304,14 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
       if (marker == 0x00 || marker == 0xFF) { base_pos += 2; continue; }
       if (marker >= 0xD0 && marker <= 0xD7) { base_pos += 2; continue; }
       if (marker == 0xD9) break;      // EOI — shouldn't happen here but guard
+      // OOB check: need at least 2 bytes for the segment length field
+      if (base_pos + 4 > base_size) break;
       int seg_len = (base_data[base_pos+2] << 8) | base_data[base_pos+3];
+      // OOB check: the segment must fit within the remaining data
+      if (base_pos + 2 + static_cast<size_t>(seg_len) > base_size) break;
+      // Skip APP markers (0xE0-0xEF) to avoid duplicating metadata
+      // (JFIF/XMP/ICC/ISO) that was already written above
+      if (marker >= 0xE0 && marker <= 0xEF) { base_pos += 2 + seg_len; continue; }
       UHDR_ERR_CHECK(Write(dest, &base_data[base_pos], 2 + seg_len, pos));
       base_pos += 2 + seg_len;
   }
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -1207,6 +1207,16 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
   UHDR_ERR_CHECK(Write(dest, &photos_editing_formats::image_io::JpegMarker::kStart, 1, pos));
   UHDR_ERR_CHECK(Write(dest, &photos_editing_formats::image_io::JpegMarker::kSOI, 1, pos));
 
+  // ── Extract APP0(JFIF) from base JPEG and write it first (per JFIF spec, APP0 must be first) ──
+  {
+    uint8_t* base_data = (uint8_t*)final_primary_jpg_image_ptr->data;
+    size_t base_size = final_primary_jpg_image_ptr->data_sz;
+    if (base_size > 4 && base_data[2] == 0xFF && base_data[3] == 0xE0) {
+        int jfif_len = (base_data[4] << 8) | base_data[5];
+        UHDR_ERR_CHECK(Write(dest, &base_data[2], 2 + jfif_len, pos));
+    }
+  }
+
   // Write EXIF
   if (pExif != nullptr) {
     const size_t length = 2 + pExif->data_sz;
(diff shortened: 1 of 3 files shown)


lib/src/jpegrutils.cpp
```diff
@@ -424,6 +424,7 @@ const string kMapOffsetHdr = Name(kGainMapPrefix, "OffsetHDR");
 const string kMapHDRCapacityMin = Name(kGainMapPrefix, "HDRCapacityMin");
 const string kMapHDRCapacityMax = Name(kGainMapPrefix, "HDRCapacityMax");
 const string kMapBaseRenditionIsHDR = Name(kGainMapPrefix, "BaseRenditionIsHDR");
+const string kMapOplusScale = Name(kGainMapPrefix, "OplusScale");
```
DichenZhang1: I didn't see this as the standard in https://developer.android.com/media/platform/hdr-image-format#example_gain_map_XMP.

minatoAI: Thank you for the review! Let me clarify the two XMP changes separately:

1. rdf:about=""
This attribute is actually present in the official Android Ultra HDR spec example: https://developer.android.com/media/platform/hdr-image-format#example_gain_map_XMP . The example shows <rdf:Description rdf:about="" ...>. Including it ensures better compatibility with strict XMP parsers.

2. hdrgm:OplusScale="-1"
You're right — this field is NOT part of the Android Ultra HDR specification. I added it because some real-world camera-captured Ultra HDR images (e.g., from OPPO devices) include this field. However, since the goal of this PR is to align with the standard, I will remove OplusScale from the XMP output. If needed, this vendor-specific extension can be revisited separately.

lib/src/jpegr.cpp
```diff
@@ -1262,12 +1272,48 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
...
+
+  // ── Skip APP0(JFIF) in base JPEG data (already written above) ──
+  if (base_pos + 4 <= base_size && base_data[base_pos] == 0xFF && base_data[base_pos+1] == 0xE0) {
+      int jfif_len = (base_data[base_pos+2] << 8) | base_data[base_pos+3];
+      base_pos += 2 + jfif_len;
+  }
+
+  // ── Write marker segments (DQT, SOF, DHT) up to SOS ──
+  size_t sos_offset = 0;
+  while (base_pos < base_size) {
```
DichenZhang1, line 1308: I think the while loop has the risk of duplicating existing APP markers (EXIF/XMP/...), maybe we need to add code
```
      if (marker >= 0xE0 && marker <= 0xEF) {
          base_pos += 2 + seg_len;
          continue;
      }
```

Also it has risk of OOB for seg_len, we'd better add a boundary check
```
      if (base_pos + 4 > base_size) {
          return status_invalid_param; // Bounds check before reading segment length
      }
      int seg_len = (base_data[base_pos+2] << 8) | base_data[base_pos+3];
      if (base_pos + 2 + static_cast<size_t>(seg_len) > base_size) {
          return status_invalid_param; // Bounds check before copying segment
      }
```

DichenZhang1 (changes requested):

lib/src/jpegr.cpp
minatoAI: Thank you for the thorough review! I've addressed both issues:

**1. APP marker duplication** — Added a check to skip APP markers (0xE0–0xEF) in the parsing loop. Since JFIF/XMP/ICC/ISO have already been written to the output earlier in the function, any APP markers encountered when walking the base JPEG are now safely skipped with `continue`:

```cpp
if (marker >= 0xE0 && marker <= 0xEF) { base_pos += 2 + seg_len; continue; }
```

**2. OOB bounds checks for `seg_len`** — Added three boundary guards in the while loop:

- Before reading the marker byte: `if (base_pos + 1 >= base_size) break;`
- Before reading `seg_len` (need 2 bytes for the length field): `if (base_pos + 4 > base_size) break;`
- After reading `seg_len`, ensuring the segment fits within remaining data: `if (base_pos + 2 + static_cast<size_t>(seg_len) > base_size) break;`

These prevent out-of-bounds reads on malformed or truncated base JPEG input. The loop uses `break` (not `continue`) for OOB conditions, consistent with the existing error-handling style in the function.

PTAL when you have a chance — thanks!

lib/src/jpegr.cpp
```diff
@@ -1262,12 +1272,56 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
...
+      if (marker >= 0xD0 && marker <= 0xD7) { base_pos += 2; continue; }
+      if (marker == 0xD9) break;      // EOI — shouldn't happen here but guard
+      // OOB check: need at least 2 bytes for the segment length field
+      if (base_pos + 4 > base_size) break;
+      int seg_len = (base_data[base_pos+2] << 8) | base_data[base_pos+3];
+      // OOB check: the segment must fit within the remaining data
+      if (base_pos + 2 + static_cast<size_t>(seg_len) > base_size) break;
+      // Skip APP markers (0xE0-0xEF) to avoid duplicating metadata
+      // (JFIF/XMP/ICC/ISO) that was already written above
+      if (marker >= 0xE0 && marker <= 0xEF) { base_pos += 2 + seg_len; continue; }
```
DichenZhang1, line 1326: There is a risk if primary JPEG contains ICC, the pointer pICC will be null, and ICC will be skipped here. Maybe we need to extract ICC from the primary JPEG first.

e.g. around line 1200, add
```
if (pIcc == nullptr && decoder.getICCSize() > 0) {
   pIcc = decoder.getICCPtr();
   icc_size = decoder.getICCSize();
 }
```

Then at line 1246 the ICC block will be written, and at here the ICC block will be ignored to prevent duplication.

DichenZhang1 (changes requested):

lib/src/jpegr.cpp
minatoAI: Thanks for reviewing! Applied exactly as suggested in the latest commit — right before the "Write ICC" block in `appendGainMap()`:

```cpp
// If the primary JPEG already carries an ICC profile and none was passed in, reuse it so the
// "Write ICC" block below emits it; the APP marker skip in the reorder loop would otherwise
// silently drop it from the output.
if (pIcc == nullptr && decoder.getICCSize() > 0) {
  pIcc = decoder.getICCPtr();
  icc_size = decoder.getICCSize();
}
```

Two points worth noting on why this is safe to write out as-is:

1. **Format compatibility.** `JpegDecoderHelper` stores the ICC marker via `jpeg_extract_marker_payload()`, which copies the *full* APP2 payload — including the `ICC_PROFILE\0` signature and the chunk index bytes (see `jpegdecoderhelper.cpp`). That is byte-identical to the layout `IccHelper::writeIccProfile()` produces, so the existing "Write ICC" block emits it unchanged, and the output ends up with exactly one ICC copy.

2. **Affected paths.** This covers the encode paths where the profile lives only inside the compressed base JPEG and the caller passes `icc = nullptr` — e.g. `encodeJPEGR()` re-encoding an SDR JPEG, where before this fix the reorder loop's APP-marker skip (`0xE0`–`0xEF`) silently dropped the profile.

Verified at runtime: the encoded output's marker sequence is `SOI → APP0(JFIF) → APP2(ICC) → APP2(ISO 21496-1) → DQT → SOF0 → DHT → APP2(MPF) → SOS`, with a single ICC segment in the primary image, and the file decodes back cleanly.

While auditing this path I also found and fixed two issues introduced by the segment reorder itself (follow-up commit `55f5248`):

- **MPF primary image size off-by-2.** The original code wrote the primary image body with `data_sz - 2` (skipping SOI), which happened to cancel out the 2-byte MPF APP2 marker; the reordered path writes exactly `base_size - sos_offset` bytes, so the marker bytes were no longer accounted for. `primary_image_size` is now `pos + 2 + length + (base_size - sos_offset)`, and the MP entries were byte-verified against the actual secondary-image SOI offset in the output.

- **Missing-SOS handling.** If a malformed base JPEG yields `sos_offset == 0`, the function now returns `UHDR_CODEC_INVALID_PARAM` with a descriptive detail message instead of silently emitting a truncated file.

PTAL, thanks!

DichenZhang1 (approved):
Thanks for catching and tackling this marker ordering issue! The updates cleanly resolve the ICC profile preservation, and the extra safety guards are a great touch. Verified end-to-end with real images—everything looks solid and spec-compliant. Thanks for your work on this!

PR #399, opened 2026-06-16
Add support for libsmpte2094-50

vrabaud:


.github/workflows/cmake_linux.yml
```diff
@@ -35,6 +35,14 @@ jobs:
             cxx: g++
             cmake-opts: '-DUHDR_BUILD_TESTS=1 -DUHDR_ENABLE_LOGS=1 -DUHDR_BUILD_DEPS=1 -DUHDR_ENABLE_WERROR=1'
 
+          # <Ubuntu-latest Platform, Release Build, GCC toolchain, Ninja generator, Build Deps, libsmpte2094_50>
+          - name: "ubuntu latest gcc rel ninja with deps and libsmpte2094_50"
+            os: ubuntu-latest
+            build_type: Release
+            cc: gcc
+            cxx: g++
+            cmake-opts: '-DUHDR_BUILD_TESTS=1 -DUHDR_ENABLE_LOGS=1 -DUHDR_BUILD_DEPS=1 -DUHDR_ENABLE_WERROR=1 -D=UHDR_ENABLE_SMPTE2094_50=1'
```
DichenZhang1: Is this a typo?

`-D=UHDR_ENABLE_SMPTE2094_50=1` should be `-DUHDR_ENABLE_SMPTE2094_50=1`

lib/src/gainmapmath.cpp
```diff
@@ -1683,4 +1687,54 @@ bool floatToUnsignedFraction(float v, uint32_t* numerator, uint32_t* denominator
...
+    }
+  }
+
+  if (isSingleChannel) {
+    delete[] memory[1];
+    delete[] memory[2];
+    memory[1] = memory[2] = nullptr;
+    mGainTable[1] = mGainTable[2] = mGainTable[0];
+  } else {
+    // Currently, assuming 1 channel if not implemented, else would need to parse rules for 3 channels.
```
DichenZhang1: Are we intended to only support single channel case? Is multiple channels case still valid? Is it better to fall back to single channel if we want to support multiple channels in the future rather than doing nothing here?

.github/workflows/cmake_linux.yml
vrabaud: done

DichenZhang1: Is this updated? I didn't see it on my side, please double check. Thank you!

vrabaud: done now

lib/src/gainmapmath.cpp
vrabaud: done

vrabaud:
Actually, this PR loses a bit of information because of the component mix that it does not use properly. Would there be a better place for me to put code ? The input would be the original image  + AGTM metadata, the output would be a full gainmap.

DichenZhang1:
> The input would be the original image + AGTM metadata, the output would be a full gainmap.
Do you want to support it reversely, e.g. primary image + gain map (the Ultra HDR image input) -> primary image + AGTM in ICC?

I think it's better to split the logic into a standalone file agtm.h/.cpp, and this may also require API change

DichenZhang1 (approved):

PR #411, opened 2026-07-30
Add support for powerpc

marillat:


kleisauke:
See PR #383 for an alternative solution.

marillat:
Of course a better solution.

PR #414, opened 2026-08-07
Check JPEG_FOUND before finding the JPEG package

SenorBlanco:
Check for JPEG_FOUND before calling find_package(JPEG). This allows a project which depends libjpeg-turbo already via add_subdirectory() to share that implementation with libultrahdr.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -423,7 +423,7 @@ if(UHDR_ENABLE_GLES)
 endif()
 
 # libjpeg-turbo
-if(NOT UHDR_BUILD_DEPS)
+if(NOT UHDR_BUILD_DEPS AND NOT JPEG_FOUND)
   find_package(JPEG QUIET)
   if(NOT JPEG_FOUND)
     message(FATAL_ERROR "Could NOT find JPEG (missing: JPEG_LIBRARIES JPEG_INCLUDE_DIRS),\


ram-mohan:
If UHDR_BUILD_DEPS is set then the project clones and Build project dependencies and not use pre-installed packages. This is not the correct change.

SenorBlanco:
Here's the problem: I already have libjpeg-turbo in my project (a particular version added via add_subdirectory()), and I don't want libultrahdr to download and build another copy, since the symbols will conflict. I also don't want libultrahdr to use FindPackage to find it (since it may not be available, e.g. on Windows) or may be the wrong version. I just want libultrahdr to use the version I'm already linking against. I can achieve this by setting JPEG_LIBRARIES, JPEG_INCLUDE_DIRS and by setting JPEG_FOUND.

If there's another alternative, I'm happy to try it.

kmilos:
> a particular version added via add_subdirectory()
>
>I also don't want libultrahdr to use FindPackage
>
> If there's another alternative, I'm happy to try it.

~~Try pointing the existing [`find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html) to your local, particular libjpg-turbo version by passing `JPEG_ROOT` (or `JPEG_DIR`), they should have priority over the default search paths? (Might need to force CONFIG mode though...)~~

Edit: Forget it, and sorry for any wild goose chase - libjpeg-turbo in config mode needs to be changed to `find_package(libjpeg-turbo CONFIG)` and defines a `libjpeg-turbo::jpeg` target, not a drop-in replacement for `JPEG::JPEG`...

However, pointing [`JPEG_INCLUDE_DIR`](https://gitlab.kitware.com/cmake/cmake/-/blob/v4.4.2/Modules/FindJPEG.cmake#L92) (not `JPEG_INCLUDE_DIRS`) and [`JPEG_LIBRARY`](https://gitlab.kitware.com/cmake/cmake/-/blob/v4.4.2/Modules/FindJPEG.cmake#L99) (not `JPEG_LIBRARIES`) might work for regular `find_package(JPEG)`...

SenorBlanco:
> ~Try pointing the existing [`find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html) to your local, particular libjpg-turbo version by passing `JPEG_ROOT` (or `JPEG_DIR`), they should have priority over the default search paths? (Might need to force CONFIG mode though...)~
> 
> Edit: Forget it, and sorry for any wild goose chase - libjpeg-turbo in config mode needs to be changed to `find_package(libjpeg-turbo CONFIG)` and defines a `libjpeg-turbo::jpeg` target, not a drop-in replacement for `JPEG::JPEG`...

No worries! I appreciate the help.

> However, pointing [`JPEG_INCLUDE_DIR`](https://gitlab.kitware.com/cmake/cmake/-/blob/v4.4.2/Modules/FindJPEG.cmake#L92) (not `JPEG_INCLUDE_DIRS`) and [`JPEG_LIBRARY`](https://gitlab.kitware.com/cmake/cmake/-/blob/v4.4.2/Modules/FindJPEG.cmake#L99) (not `JPEG_LIBRARIES`) might work for regular `find_package(JPEG)`...

Setting those is enough to satisfy `find_package()`, but unfortunately the compile fails with

```
FAILED: third_party/libultrahdr/CMakeFiles/core.dir/lib/src/editorhelper.cpp.obj
C:\PROGRA~1\MICROS~1\2022\COMMUN~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\cl.exe  /nologo /TP -DSTACK_SIZE=4194304 -D_CRT_SECURE_NO_WARNINGS -IC:\Users\Stephen\src\toucan\third_party\libultrahdr\lib\include -I\third_party\libjpeg-turbo -IC:\Users\Stephen\src\toucan\third_party\libultrahdr\third_party\image_io\includes -IC:\Users\Stephen\src\toucan\third_party\libultrahdr /DWIN32 /D_WINDOWS /EHsc /O2 /Ob2 /DNDEBUG -std:c++17 -MD -DUHDR_ENABLE_INTRINSICS -DUHDR_WRITE_ISO /wd4244 /wd4267 /wd4305 /wd4838 /wd26812 /showIncludes /Fothird_party\libultrahdr\CMakeFiles\core.dir\lib\src\editorhelper.cpp.obj /Fdthird_party\libultrahdr\CMakeFiles\core.dir\core.pdb /FS -c C:\Users\Stephen\src\toucan\third_party\libultrahdr\lib\src\editorhelper.cpp
C:\Users\Stephen\src\toucan\third_party\libultrahdr\lib\include\ultrahdr/jpegdecoderhelper.h(27): fatal error C1083: Cannot open include file: 'jerror.h': No such file or directory
```

I suspect because JPEG clients need both `jerror.h` (from the source directory) and `jconfig.h` (from the binary directory). I think this is why libultrahdr's does this when building libjpegturbo with `UHDR_BUILD_DEPS`:

```
set(JPEG_INCLUDE_DIRS ${JPEGTURBO_SOURCE_DIR} ${JPEGTURBO_BINARY_DIR})
```

kmilos:
> I suspect because JPEG clients need both jerror.h (from the source directory) and jconfig.h (from the binary directory).

Sounds like your local libjpeg-turbo wasn't **_installed_** properly _out of its source tree_ (extra step after just building it)? No problems w/ compiling w/ system libjpeg-turbo packages in Linux distros or e.g. MSYS2 for Windows. All the header files should get _installed_ in the same location, see e.g. file listings in

https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-libjpeg-turbo
https://archlinux.org/packages/extra/x86_64/libjpeg-turbo/files/
https://packages.fedoraproject.org/pkgs/libjpeg-turbo/libjpeg-turbo-devel/fedora-44.html#files

etc.

>  I think this is why libultrahdr's does this when building libjpegturbo with UHDR_BUILD_DEPS

Bundling dependencies and building everything in-tree probably requires many custom workarounds...

SenorBlanco:
OK, I found another solution.

I switched to using `find_package(JPEG)` in my project. If not found, I build the hermetic version of `libjpeg-turbo` in my project, set `JPEG_INCLUDE_DIR` and `JPEG_LIBRARY` to it, alias `JPEG::JPEG` to it via `add_library()` and set `JPEG_FOUND=TRUE`. This is enough to satsify the `find_package(JPEG)` call inside `libultrahdr` to always succeed.

The only downside to this approach is that building `libjpeg-turbo` is done in parallel with other targets, and when using Ninja will exceed the core count of the machine. But this is really a `CMake` problem, not specific to `libultrahdr`.

PR #443, opened 2026-08-13
gainmapmath: write green channel in RGB888 output

gregbenz:
## What changed

- Write `pixel.g` to the green byte in `putRgb888Pixel()`.
- Add a regression test using distinct red, green, and blue values.

## Why

`putRgb888Pixel()` currently writes `pixel.r` to both the red and green bytes. Any
RGB888 output where red and green differ is therefore color-corrupted. Writing the
actual green component preserves the requested RGB pixel values.

## Testing

- Added `GainMapMathTest.PutRgb888PixelWritesGreenChannel`; it fails on current main
  and passes with this change.
- Full non-HEIF unit suite: 1,291 tests run; 1,067 passed and 224 existing parameter
  combinations skipped.


DichenZhang1 (approved):
Thank you for finding and fixing this bug!

gregbenz:
@DichenZhang1 thank you and the team for all these great enhancements and ongoing support of this critical library for HDR!

PR #448, opened 2026-08-14
Fix HEIF/AVIF output buffer overflow

gregbenz:
## What changed

- Make the libheif memory writer reject size overflow and allocation failure.
- Propagate errors returned by `heif_context_write()`.
- Reject encoded output larger than the caller's destination capacity before copying it.
- Preserve the destination's capacity and add HEIF/AVIF regression tests.

## Why

The HEIF and AVIF encoders copied the complete encoded file into the destination without checking its capacity. An undersized destination could therefore be overwritten while the encoder reported success. The encoder now returns `UHDR_CODEC_MEM_ERROR` without modifying the destination buffer.

## Testing

- Both regressions fail on current main: capacity 1 accepts and copies 15,245 bytes for HEIF and 2,323 bytes for AVIF.
- Both regressions pass with this fix.
- Full HEIF-enabled unit suite: 1,072 passed, 224 skipped, 0 failed.
- Both modified codec translation units compile with `-fno-exceptions` and warnings as errors.

### DIFF
--- a/lib/src/avifultrahdr.cpp
+++ b/lib/src/avifultrahdr.cpp
@@ -17,6 +17,7 @@
 #ifdef UHDR_ENABLE_HEIF
 
 #include <cstdlib>
+#include <limits>
 #include <map>
 
 #include "ultrahdr/avifultrahdr.h"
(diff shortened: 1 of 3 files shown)


gregbenz:
@DichenZhang1 FYI that this was flagged by the google-cla check, which is pretty odd as I've done that, contributed, and other PRs sent yesterday were not flagged. Not sure why this one was, but seems incorrectly flagged by that bot. Perhaps that automatic build was a factor? It appears to have been a transient build/test issue (Error: socket hang up)

DichenZhang1 (approved):

Issue #379, opened 2026-01-18
ultrahdr_app throws an error when trying to write a path with Chinese characters


gregbenz:
It appears the CLI tool cannot handle some valid characters in paths, such as the Chinese character below.

Using the command line flag:
`-z \"C:\Users\admin\Documents\張\out.jpg"`

Will show an error message:
**unable to write to file: C:\Users\admin\Documents\?\out.jpg**

This fails under Windows 11, but seems to work OK under MacOS.

_Note that other less common characters in EU languages will work in the CLI app (if they exist on the disk) - but if the path fails for other reasons, the error message will be mangled. For example, "À" shows as "└" in the path listed in the error. So the error message itself does not correctly list the path in some cases where the character itself is supported._

kmilos:
Perhaps forcing the [UTF-8 codepage on modern Windows 10 1903 or later](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page) is sufficient for this... Something along the lines of https://github.com/libjxl/libjxl/commit/7b70ef19383a73d8a44bd9bbe15df140412df300

PR #452, opened 2026-08-16
Support Unicode file paths in the Windows CLI

gregbenz:
## Summary

- Enable UTF-8 paths in `ultrahdr_app` through the Windows active-code-page manifest.
- Add the manifest directly for MSVC and through a resource file for MinGW.
- Add an end-to-end test using CJK and emoji input/output paths.

## Why

The Windows CLI otherwise interprets narrow command-line arguments and file paths using the legacy system code page. Valid Unicode filenames may therefore be changed or rejected.

Using the UTF-8 application manifest fixes this without custom command-line parsing or per-path conversions. This follows the same approach used by libavif.

## Compatibility

This affects only `ultrahdr_app`; the library API/ABI and non-Windows behavior are unchanged.

Full Unicode path support requires Windows 10 version 1903 or newer, including Windows 11. Earlier Windows versions can still run the CLI, but arbitrary Unicode paths remain limited by the legacy system code page.

Long-path policy and Windows-reserved filenames are outside this change.

Addresses #379.

## Validation

Google’s native Windows CI passed:

- `UHDRUnitTests`
- `UHDRWindowsUnicodeCliPaths`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -787,6 +787,14 @@ if(UHDR_BUILD_EXAMPLES)
     target_link_options(${UHDR_SAMPLE_APP} PRIVATE -static)
   endif()
   target_link_libraries(${UHDR_SAMPLE_APP} PRIVATE ${UHDR_CORE_LIB_NAME})
+  if(WIN32)
+    if(MSVC)
+      target_sources(${UHDR_SAMPLE_APP} PRIVATE "${EXAMPLES_DIR}/utf8.manifest")
+    elseif(MINGW)
+      # MinGW does not have a manifest tool (mt.exe), so wrap the manifest in a resource script.
+      target_sources(${UHDR_SAMPLE_APP} PRIVATE "${EXAMPLES_DIR}/utf8.rc")
+    endif()
+  endif()
 endif()
 
 if(UHDR_BUILD_TESTS OR UHDR_BUILD_BENCHMARK)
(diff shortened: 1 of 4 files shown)


kmilos:
As already Windows 10 is EOL, it might not be worth carrying this wide vs narrow legacy cruft around. A more elegant/modern solution is to use ultrahdr apps (and other clients of the library) in the [native UTF-8 code page](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page).

See e.g. https://github.com/AOMediaCodec/libavif/commit/3ec01cefd1ddd266a622d5e114a0888581b68f4a

gregbenz:
Thanks for the suggestion—the UTF-8 manifest is a cleaner approach for modern Windows, and the libavif precedent is helpful.

There is one compatibility tradeoff:

- The current PR explicitly converts the native UTF-16 command line and therefore supports Unicode paths across older Windows versions, including Windows 10 LTSC 2019 and Windows Server 2019.
- The manifest approach requires Windows 10 version 1903 or newer. It substantially reduces the application code, but Unicode paths would remain unsupported on older builds. LTSC 2019 and Server 2019 are based on version 1809 and remain supported until 2029.

Microsoft documents the version requirement here:
https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

@DichenZhang1 Would the maintainers prefer the simpler manifest approach and effectively require Windows 10 1903+, or retain the explicit conversion for broader Windows compatibility? I’m happy to revise the PR either way and will keep the existing Windows Unicode regression test to verify the chosen implementation.

kmilos:
> * LTSC 2019 and Server 2019 are based on version 1809 and remain supported until 2029.

Only on _mission critical_ equipment like medical devices etc. I don't think they'd be deploying new (read "not yet withstood the test of time") features and libraries like libultrahdr on those.

gregbenz:
@kmilos Thank you, helpful context. I like your proposal and agree it seems simpler and a good approach. I would like to hear if the maintainers have a preference, I'm happy to revise the approach but just want to confirm direction.

DichenZhang1 (approved):

PR #413, opened 2026-08-03
Add HEIF (HEIC) and AVIF Ultra HDR multi-format container support (v2.0.0)

DichenZhang1:
## Overview

This pull request introduces multi-format container support for **Ultra HDR v2.0.0**, extending `libultrahdr` beyond JPEG to support **HEIF/HEIC (`.heic`)** and **AVIF (`.avif`)** images with ISO 21496-1 gain map metadata and color profile preservation.

---

### Key Architectural Changes

1. **Modular `UltraHdr` Base Class Hierarchy**:
   - Refactored the monolithic `JpegR` implementation by extracting core gain map algorithms into a shared `UltraHdr` base class (`ultrahdrcommon.h`).
   - Added specialized derived classes:
     - `JpegR` (`jpegr.h` / `jpegr.cpp`): JPEG Multi-Picture Format (MPF), EXIF, and XMP encapsulation.
     - `HeifUltraHdr` (`heifultrahdr.h` / `heifultrahdr.cpp`): HEVC bitstream encoding/decoding via `libheif` with ISO 21496-1 `tmap` tone-mapping item references.
     - `AvifUltraHdr` (`avifultrahdr.h` / `avifultrahdr.cpp`): AV1 bitstream encoding/decoding via `libheif` with ISO 21496-1 `tmap` item references.

2. **C-API Multi-Format Extensions**:
   - Activated `UHDR_CODEC_HEIF` and `UHDR_CODEC_AVIF` in `uhdr_enc_set_output_format()`.
   - Enhanced `uhdr_dec_probe()` with automatic container sniffing (ISO BMFF `ftyp` vs JPEG SOI markers) and metadata extraction.
   - Transparently dispatched `uhdr_decode()` across JPEG, HEIF, and AVIF streams.

3. **Color Space & Gain Map Conversion Robustness**:
   - Extracted NCLX container properties directly from image handles with fallbacks to standard color spaces.
   - Added RGBA-to-monochrome gain map pixel conversion in `copy_raw_image()` for decoders producing RGB/RGBA chroma formats.

4. **Testing, Portability & Documentation**:
   - Added comprehensive end-to-end unit test suite in `tests/ultrahdr_api_test.cpp` covering all API entry points.
   - Cross-platform CMake configuration supporting standalone builds on Linux, macOS (Intel & Apple Silicon), and Windows.
   - Documented all architectural changes in `docs/api_changes_v1.5_to_v2.0.md`.
[...]

Issue #391, opened 2026-05-18
Two heap OOB reads in encoder pixel processing on odd-dimension images


jortles:
# Heap OOB reads in encoder pixel-processing on odd dimensions

## Summary

Two heap-buffer-overflow READ bugs exist in libultrahdr's internal pixel-processing functions when image dimensions are odd. The public encoder API validates dimensions (rejecting odd widths/heights for P010 and YCbCr420), so these are not reachable through normal API usage. The decode path uses different pixel accessors that are not affected. These are defense-in-depth issues where internal functions lack bounds checking and rely entirely on caller validation.

## Bug 1: getP010Pixel UV overflow on odd width

**File**: `lib/src/gainmapmath.cpp:443`

P010 chroma is interleaved U,V pairs. The V index is computed as `pixel_u_idx + 1`. For odd width `w` at the last pixel in a row (`x = w-1`, which is even):
- `pixel_v_idx = (y/2) * w + w` -- exactly 1 element past the chroma buffer end.

**OOB**: 2-byte read (one uint16_t) past the allocated chroma plane.

### ASAN trace

```
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x524000005df2
READ of size 2 at 0x524000005df2 thread T1
    #0 ultrahdr::getP010Pixel(uhdr_raw_image*, unsigned long, unsigned long) gainmapmath.cpp:443
    #1 ultrahdr::samplePixels(...) gainmapmath.cpp:505
    #2 ultrahdr::sampleP010(...) gainmapmath.cpp:525
    #3 ultrahdr::JpegR::generateGainMap(...)::$_0::operator()() jpegr.cpp:778
0x524000005df2 is located 0 bytes after 7410-byte region [0x524000004100,0x524000005df2)
```

## Bug 2: convert_raw_input_to_ycbcr row overflow on odd height

**File**: `lib/src/gainmapmath.cpp:1337` (RGBA1010102) and `:1431` (RGBA8888)

The function processes pixels in 2x2 blocks. For odd height `h`, the last iteration (`i = h-1`, even) accesses row `i+1 = h` -- one row past the buffer.

**OOB**: 4-byte read (one uint32_t) past the source image allocation.

### ASAN trace

```
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x525000002740
READ of size 4 at 0x525000002740 thread T0
[...]

PR #459, opened 2026-08-20
gainmapmath: restore odd-dimension OOB read fixes reverted by #413

MajorMurphy:
Restores the defect fix from #391, originally submitted by @jortles in #392 and landed by @DichenZhang1 as #407 (`586ed42`).

#413's `07bc6a2` was branched before #407 and reverted all three hunks when it merged nine days later, so v2.0.0, v2.0.1, v2.0.2 and current `main` all ship the pre-fix code:

- **`lib/src/gainmapmath.cpp:432`** — `getP010Pixel`'s chroma index is back to `(x & ~0x1)` with no clamp, so `pixel_v_idx = pixel_u_idx + 1` reads one element past the chroma plane on the last column of an odd-width P010 image.
- **`lib/src/gainmapmath.cpp:1320` and `:1413`** — `convert_raw_input_to_ycbcr`'s 2×2 block loops are back to `i < dst->h` / `j < dst->w`, reading row `i + 1` / column `j + 1` past the edge on odd dimensions. Both the RGBA1010102 and RGBA8888 paths.

`git diff 586ed42 07bc6a2 -- lib/src/gainmapmath.cpp` shows the three hunks being undone. The revert looks incidental rather than deliberate — the HEIF/AVIF staging branch predated #407 and carried the older copy of the file forward over it.

#407's test `GetP010PixelOddWidth` was *not* reverted and is still in `tests/gainmapmath_test.cpp`, but it asserts `EXPECT_NO_FATAL_FAILURE`, which a short read past a heap allocation does not trip outside a sanitizer build. The test and the code it guards were reverted independently, leaving a green guard over absent code.

This is a cherry-pick of `586ed42` onto current `main`, preserving its original authorship. The test hunk is already present, so only `lib/src/gainmapmath.cpp` changes (+9 / −5).

One caveat carried over from #407: the loop-bound half skips the trailing incomplete block rather than clamping, so the final row and column of an odd-dimension image are left value-initialized — black with fully saturated chroma — instead of read out of bounds. Clamping the source index the way `getP010Pixel` now does would fix the values too, but I have kept this to exactly what #407 did so it reads as a pure restoration. Happy to follow up.

Verified to build clean on MSVC 2022 x64 Release; I have not run it under ASan.

### DIFF
--- a/lib/src/gainmapmath.cpp
+++ b/lib/src/gainmapmath.cpp
@@ -429,7 +429,11 @@ Color getP010Pixel(uhdr_raw_image_t* image, size_t x, size_t y) {
   size_t chroma_stride = image->stride[UHDR_PLANE_UV];
 
   size_t pixel_y_idx = y * luma_stride + x;
-  size_t pixel_u_idx = (y >> 1) * chroma_stride + (x & ~0x1);
+  size_t chroma_x = x & ~(size_t)0x1;
+  if (chroma_x + 1 >= image->w) {
+    chroma_x = (image->w >= 2) ? ((image->w - 2) & ~(size_t)0x1) : 0;
+  }
+  size_t pixel_u_idx = (y >> 1) * chroma_stride + chroma_x;
   size_t pixel_v_idx = pixel_u_idx + 1;
 
   uint16_t y_uint = luma_data[pixel_y_idx] >> 6;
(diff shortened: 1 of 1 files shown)


gregbenz:
Comparing this against #407 and #413, it does restore all three OOB protections that were lost. The caveat in the description is accurate: changing the loop bounds prevents invalid reads, but leaves any trailing odd row or column unwritten.

A complete fix would continue processing partial 2×2 blocks, replicate edge samples for chroma averaging, and write luma only at valid coordinates. It also requires ceil-divided 4:2:0 geometry through allocation, copying, and the AVIF/HEIF handoff. That is larger than a safe inline suggestion and seems better scoped as a follow-up.

I think this narrow safety restoration is worth merging as-is. I have a locally prepared and tested a potential follow-up covering 1×N, N×1, odd-width, odd-height, and both-odd RGB888, RGBA8888, and RGBA1010102/P010 cases, plus an AVIF codec regression. It will need to be rebased after #459 lands, and I can prepare a separate follow-up PR then.

DichenZhang1 (approved):

PR #467, opened 2026-09-11
Write ISO gain-map metadata with independent fractions

gregbenz:
When all stored denominators match, the ISO gain-map metadata writer sets reserved bit 3 and emits the older compact layout. This is reachable through the supported JPEG API-4 path: one- and three-channel metadata produce 37- and 77-byte payloads instead of the full 61- and 141-byte layout.

This change always writes each numerator and denominator independently, matching [libavif's layout update](https://github.com/AOMediaCodec/libavif/commit/669bea7e5bdc089e864ab8da94760270a80875cd). It also rejects the unsupported internal `backwardDirection` before writing. Current ISO metadata represents direction through the base and alternate HDR headroom values rather than a flag, so all six reserved bits remain zero. The shared writer is used by JPEG, AVIF, and HEIF; decoding of existing compact metadata is unchanged.

Tests add exact one- and three-channel byte vectors for equal and mixed denominators, fixed legacy compact decoder vectors, and checks that backward-direction rejection does not modify the output.

Local validation:

- strict no-HEIF core build
- focused metadata tests: 9 passed
- focused ASan/UBSan tests: 9 passed
- full no-HEIF suite: 1,078 passed, 224 existing skips
- full-size JPEG API-4 checks produced and decoded 61- and 141-byte metadata payloads
- full-size sRGB- and Display-P3-base AVIF outputs, with one- and three-channel gain maps, were accepted by libultrahdr and `avifgainmaputil`

This PR does not add HDR-base encoding, tighten reserved-bit parsing, remove legacy input support, or change container topology.

### DIFF
--- a/lib/src/gainmapmetadata.cpp
+++ b/lib/src/gainmapmetadata.cpp
@@ -122,15 +122,20 @@ uhdr_error_info_t uhdr_gainmap_metadata_frac::encodeGainmapMetadata(
     return status;
   }
 
+  if (in_metadata->backwardDirection) {
+    uhdr_error_info_t status;
+    status.error_code = UHDR_CODEC_UNSUPPORTED_FEATURE;
+    status.has_detail = 1;
+    snprintf(status.detail, sizeof status.detail, "backward direction is not supported");
+    return status;
+  }
+
   const uint16_t min_version = 0, writer_version = 0;
   streamWriteU16(out_data, min_version);
   streamWriteU16(out_data, writer_version);
 
   uint8_t flags = 0u;
-  // Always write three channels for now for simplicity.
-  // TODO(maryla): the draft says that this specifies the count of channels of the
-  // gain map. But tone mapping is done in RGB space so there are always three
-  // channels, even if the gain map is grayscale. Should this be revised?
+  // Version 0 defines only the channel-count and base-color-space flags.
   const uint8_t channelCount = in_metadata->allChannelsIdentical() ? 1u : 3u;
 
   if (channelCount == 3) {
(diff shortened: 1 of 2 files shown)


DichenZhang1 (approved):
TODO: will be addressed in PR #473

Follow-up hardening for ISO/IEC 21496-1 gain map metadata serialization and deserialization (following PR #467):
1. **Zero-Denominator Defense**:
   - In `encodeGainmapMetadata`, explicitly validate that all rational denominators (`baseHdrHeadroomD`, `alternateHdrHeadroomD`, per-channel `gainMapMinD`, `gainMapMaxD`, `gainMapGammaD`, `baseOffsetD`, `alternateOffsetD`) are non-zero before serialization, returning `UHDR_CODEC_INVALID_PARAM` if 0 is encountered.
   - In `decodeGainmapMetadata`, explicitly validate that `commonDenominator != 0` when reading legacy bitstreams.
2. **ISO/IEC 21496-1 Compliance**:
   - Strictly clear reserved bit 3 of flags (`flags & 0x08 == 0`) and serialize independent `(numerator, denominator)` fraction pairs for all rational fields.
3. **Backward Compatibility**:
   - `decodeGainmapMetadata` retains support for decoding legacy bitstreams where bit 3 is set.
## Tests Added
- `GainMapMetadataTest.EncodedFlagsClearsReservedBit3`: Verifies that emitted flags byte strictly clears reserved bit 3.
- `GainMapMetadataTest.RejectsZeroDenominatorDuringEncoding`: Asserts that any input fraction with denominator 0 is rejected with `UHDR_CODEC_INVALID_PARAM`.
- `GainMapMetadataTest.DecodesLegacyCommonDenominatorStream`: Verifies that synthetic legacy payloads with bit 3 set decode cleanly into expected headroom and channel values.

PR #471, opened 2026-09-12
Handle JPEG XMP size limits without losing metadata

gregbenz:
A valid 65,504-byte XMP packet can exceed the image-based output allocation, while a larger packet is silently omitted with a successful encode. Reserve room for one standard XMP APP1 segment when explicit or inherited XMP may be written, and reject a final serialized packet that exceeds the segment limit, including after dual-mode merging.

Adds three regression tests covering raw-image and compressed base/gain-map encoding, metadata extraction and decoding, and oversized output. All three fail against the original feature branch in both configurations; all 12 API tests pass with the fix in ISO and dual configurations.

Based on `feat/preserve-xmp-metadata` at `59418fb`. Extended XMP and XML merging behavior are outside this patch.

### DIFF
--- a/lib/src/jpegr.cpp
+++ b/lib/src/jpegr.cpp
@@ -1286,17 +1286,26 @@ uhdr_error_info_t JpegR::appendGainMap(uhdr_compressed_image_t* sdr_intent_compr
   }
 
   if (!xmp_primary_str.empty()) {
-    const size_t length = 2 + xmpNameSpaceLength + xmp_primary_str.size();
-    if (length <= 65535) {
-      const uint8_t lengthH = ((length >> 8) & 0xff);
-      const uint8_t lengthL = (length & 0xff);
-      UHDR_ERR_CHECK(Write(dest, &photos_editing_formats::image_io::JpegMarker::kStart, 1, pos));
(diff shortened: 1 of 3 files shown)


DichenZhang1:
Hi Greg,

Thank you so much for the thorough review, catching the boundary/sizing edge cases, and submitting this PR!
We've pulled your commit into `feat/preserve-xmp-metadata` and did a quick follow-up refactoring:
- Consolidated the segment length constants (`kJpegSegmentMaxLength = 0xFFFF` and `kJpegAppSegmentTotalMaxBytes = kJpegSegmentMaxLength + 2`) into `lib/include/ultrahdr/jpegr.h` so `jpegr.cpp` and `ultrahdr_api.cpp` share a single source of truth.
- Cleaned up a minor duplicate `<vector>` include in the test file.
All unit tests (including your new `LargeXmpFitsSmallRawApiOutput`, `LargeXmpFitsCompressedBaseGainmapApiOutput`, and `OversizedFinalXmpIsRejected` test cases) pass cleanly.

We really appreciate your continued expertise and contributions to libultrahdr!

gregbenz:
@DichenZhang1 Thank you for all you and the team are doing for the community!

